PR #7, opened 2022-11-10
Add search metrics, sponsored tiles metrics

skahmann3:


definitions/fenix.toml
```diff
@@ -178,6 +157,163 @@ category = "performance"
...
 
 
+#### search metrics
+
+[metrics.tagged_sap_searches]
+select_expression = "{{agg_sum('tagged_sap')}}"
+data_source = "mobile_search_clients_engines_sources_daily"
+friendly_name = "Tagged SAP Searches"
+description = "Total number of tagged SAP searches. This metric is deprecated, use tagged_search_count instead"
+deprecated = true
```
scholtzan, line 167: We can't remove this metric atm since it's still in use by some experiments. But we should mark it as deprecated

scholtzan (approved):
Thanks! lgtm

definitions/fenix.toml
xluo-ds: @skahmann3  ^^

PR #14, opened 2022-12-14
Add `is_default_browser`, `is_default_pdf_handler`, and `is_pinned` F…

ncalexan:
…irefox Desktop metrics.

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -831,6 +831,32 @@ description = "Total number of Organic Ad Click Counts"
 category = "search"
 type = "scalar"
 
+[metrics.is_default_browser]
+data_source = "clients_daily"
+select_expression = "COALESCE(LOGICAL_OR(is_default_browser), FALSE)"
+friendly_name = "Is Default Browser"
+type = "scalar"
+description = """
+Was Firefox the default browser at any point during the interval?
+"""
+
+[metrics.is_default_pdf_handler]
+data_source = "main"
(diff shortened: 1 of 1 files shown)


ncalexan:
I cribbed these from @danielkberry and extended to the default PDF handler, see https://github.com/mozilla/jetstream-config/pull/249.

danielkberry (approved):
Default browser and pinned to taskbar LGTM. I'm not very familiar with the pdf handler telemetry, but this seems reasonable? @irrationalagent  may know more. I will say: is there any place that default pdf handler status is aggregated to the (client, day) level? Scanning the main ping table can be expensive...

ncalexan:
> Default browser and pinned to taskbar LGTM. I'm not very familiar with the pdf handler telemetry, but this seems reasonable? @irrationalagent may know more.

I'm not aware of anybody who's done anything with this besides myself and Romain; see, for example, https://sql.telemetry.mozilla.org/queries/86197/source?p_channel=release.  (That's where I got this snippet from.)

> I will say: is there any place that default pdf handler status is aggregated to the (client, day) level? Scanning the main ping table can be expensive...

I looked, but I'm not aware of any such place.  Definitely it's not aggregated into `telemetry.clients_daily` or similar.  Since  the PDF outcome is not default, this seems reasonable.

ncalexan:
@danielkberry I don't have write access so could you merge when you're satisfied?  Thanks!

danielkberry:
Cool, thanks for looking into those comments.

PR #17, opened 2023-01-18
Add country segments

rebecca-burwei:
After this PR is merged, we will be able to use these Segments with mozanalysis like:
```
from mozanalysis.config import ConfigLoader
pocket_countries = ConfigLoader.get_segment("pocket_countries", app_name="firefox_desktop")
```


mmccorks (approved):
LGTM

PR #18, opened 2023-01-19
Add placeholder folder for tooling-specific configs

scholtzan:
My plan is to:

1. create these folders that will contain the tool configs
2. update jetstream to merge configs from these folders and the existing jetstream-config repo
3. announce that we are changing the location of the configs
4. update docs
5. move jetstream-config into this repo
6. update jetstream to only use configs from this repo

### DIFF
--- a/jetstream/README.md
+++ b/jetstream/README.md
@@ -1,0 +1,24 @@
+# Jetstream Config
+
+> This is currently a WIP.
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
Looks good to me! Do you want to update the top level README at this point or leave it as-is until we're ready to start using these folders?

scholtzan:
I'll wait until we move the configs over

PR #25, opened 2023-01-25
Move jetstream configs

scholtzan:


mikewilli (approved):
Changes all look good to me. Should we add a note in one of the READMEs about the precedence of definitions since there are so many similar path hierarchies in one repo now? A link out to the doc that talks about it would be perfect if that exists.

scholtzan:
Added a note to the README

PR #30, opened 2023-01-30
[skip ci] Adds Binomial statistic for `is_pinned`

danielkberry:


danielkberry:
@scholtzan I think it's fine to apply this on an ongoing basis to new experiments. I don't think it's worth rerunning like what happened when `is_default_browser` was merged

What's the best way to go about doing that?

scholtzan:
You can add a `[skip ci]` to the PR message, so rerun doesn't get started. And I can trigger a script to mark all the existing experiments as "fresher" than the config file (so the rerun doesn't get triggered during the next scheduled run)

PR #35, opened 2023-02-03
Rerun only experiments changed in PR

scholtzan:
Depends on https://github.com/mozilla/jetstream/pull/1560

The reason to change this is to only rerun configs that got changed in the PR. Previously, we reran all experiments that had a changed config. This resulted in experiments getting rerun multiple times when PRs got merged right after each other before the rerun had completed.
This was especially a problem when defaults got changed since that affected a lot of experiments, costing money.

With this change experiments affected by a change in defaults will only get rerun during nightly runs. Which I think is fine

cc @danielkberry

### DIFF
--- a/.circleci/validation-config.yml
+++ b/.circleci/validation-config.yml
@@ -106,12 +106,26 @@ jobs:
     - run:
         name: Rerun jetstream
         command: |
+          changed_files=$(git diff --name-only $BASE_COMMIT..$REVISION_COMMIT -- 'jetstream/*.toml' 'jetstream/*.toml.example')
+          echo "Rerun changed files: "
+          echo $changed_files
+
+          # determine slugs of configs that got changed; filter out outcomes and defaults
+          params=""
+          for config in $changed_files; do
+            slug=${config##*/}
+            slug="${slug%.*}"
+            if [ -e jetstream/"$slug".toml ]; then
+              params+="--experiment_slug=${slug} "
+            fi
+          done
+
           # stop running instances so that they do not interfere
           uuid=`uuidgen`
           pod_identifier="jetstream-${uuid}"
           kubectl delete pod -l app=$pod_identifier
           # start a new instance
-          kubectl run $pod_identifier --image=gcr.io/moz-fx-data-experiments/jetstream -l app=$pod_identifier --restart=Never --command -- jetstream rerun-config-changed --argo --return-status --recreate-enrollments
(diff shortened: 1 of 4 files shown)


danielkberry (approved):
This is awesome, a great change to have. I was about to make some changes to the Fenix defaults so I'm glad this is going in now.

.circleci/validation-config.yml
```diff
@@ -106,12 +106,26 @@ jobs:
...
+            fi
+          done
+
           # stop running instances so that they do not interfere
           uuid=`uuidgen`
           pod_identifier="jetstream-${uuid}"
           kubectl delete pod -l app=$pod_identifier
           # start a new instance
-          kubectl run $pod_identifier --image=gcr.io/moz-fx-data-experiments/jetstream -l app=$pod_identifier --restart=Never --command -- jetstream rerun-config-changed --argo --return-status --recreate-enrollments
+          kubectl run $pod_identifier --image=gcr.io/moz-fx-data-experiments/jetstream -l app=$pod_identifier --restart=Never --command -- jetstream rerun --argo --return-status --recreate-enrollments $(echo $params)
```
mikewilli, line 128: Is there a reason to get the updated configs from the bash script above instead of through python like we do for `rerun-config-changed`? I am concerned that this could be confusing to have them defined differently. Perhaps we could instead put the new filter into `rerun-config-changed` (and add a Click flag to the command if we want to be able to toggle between the current functionality and this new filter)?

scholtzan: `rerun-config-changed` only looks at timestamps of when the tables got updated and when configs got updated. Based on these timestamps it reruns all experiments which have a config that has been updated more recently than the table. 

The CI can determine exactly which configs/experiments just got changed in the PR that got merged and only rerun them once. I guess we could implement the same in jetstream by evaluating the git history, comparing diffs between the most recent merged commit with the previous HEAD commit on main. (this gets more complicated on rebases) 

It seemed easier to implement it this way. It might be as confusing having a click flag for it instead? At least I think, users that aren't using the `jetstream` tooling and only interface through this repo shouldn't notice a difference.

mikewilli: Does it make sense to remove the `rerun-config-changed` command or do you still envision use cases for it?

scholtzan: It'll still be used in the nightly runs: https://github.com/mozilla/telemetry-airflow/blob/d97adbdf1c747d900525528865cae6eff949e4e3/dags/jetstream.py#L63

Which will pick up and rerun experiments when defaults (and outcomes?) have changed.

mikewilli (approved):
Nice! A good start towards making the reruns more efficient. And thanks for the clarification on the decision to get changed configs directly in the CI command.

PR #36, opened 2023-02-06
Analyis for Firefox update notification experiments.

ncalexan:
These experiments are at
https://experimenter.services.mozilla.com/nimbus/mr2022-backgroundtaskmessage-notification-release-1pct and
https://experimenter.services.mozilla.com/nimbus/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback and pose some novel challenges for Jetstream; see the extensive inline commenting.

The second anaylsis is a duplicate of the first analysis, with:
- appropriate dates
- the interpolated experiment slug updated in 2 places.

### DIFF
--- a/jetstream/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback.toml
+++ b/jetstream/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback.toml
@@ -1,0 +1,214 @@
+## Background messaging tasks present several wrinkles compared to regular
+## Firefox in-product (browsing) experiments.  See [Jetstream support for
+## background tasks Firefox messaging system
+## experiments](https://docs.google.com/document/d/1GeOGUmG8XpitdexW2HESZJbCNhTT-o6qe1MmBsDPDHE/edit)
+## for an overview.
+##
+## This specific analysis uses legacy client IDs as its single unique
+## identifier, answering the main open question of the document linked above.
+##
+## This analysis is complicated by being written to be more general than this
+## single experiment; it tries to accommodate future experiments by referencing
+## the experiment slug indirectly and to handle in-product changes that will
+## apply to future experiments but do not apply to this one.
+##
+## For technical reasons, we take the union of multiple queries to determine
+## enrollment; see the extensive inline comments below.  It is possible to
+## segment by the various sub-queries, but it is awkward and does not provide
+## much value (beyond exposing counts of the various segments, which is valuable
+## during development but not relevant to impact analysis).
+##
+## Note that background update telemetry can be delayed, and that re-engaging
+## with Firefox can delay that telemetry further since the background update
+## task does not run if Firefox itself is running.  Therefore, even though the
+## true enrollment period -- the time when Nimbus is enrolling clients via
+## Remote Settings -- is generally *one* week, we artificially expand the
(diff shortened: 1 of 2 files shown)


ncalexan:
@danielkberry can you review?  Nothing's changed compared to https://github.com/mozilla/jetstream-config/pull/266.

@scholtzan again, I can't request review.  What, exactly, gates being able to request review?  Can we avoid this ping-pong in some way?

jetstream/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback.toml
```diff
@@ -0,0 +1,211 @@
...
+## true enrollment period -- the time when Nimbus is enrolling clients via
+## Remote Settings -- is generally *one* week, we artificially expand the
+## enrollment period to *two* weeks for Jetstream's analysis.
+
+[experiment]
+
+start_date = "2023-01-23"
+end_date = "2023-02-20"
+enrollment_period = 14
+
```
danielkberry, line 33: I think we should add `reference_branch = "disabled"` here. Right now, the experiment is configured with `enabled` as the control branch.

danielkberry (approved):
Agree, I don't think we need to modify it much from the jetstream-config version

jetstream/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback.toml
ncalexan: I've done this.  Should my reference branch have been named something else to have this automatically?  `"control"`?

ncalexan: Follow-up: why isn't this determined from the Experimenter recipe, which I think knows about reference/control branches?

ncalexan:
@danielkberry urgh, I need re-review because of the `reference_branch` change.

ncalexan:
> > @scholtzan again, I can't request review. What, exactly, gates being able to request review? Can we avoid this ping-pong in some way?
> 
> You should be able to do so now. The reason is that for all repositories created under the mozilla organisation, any member of the org only has "Viewer" permissions by default. For requesting reviews you need "Triage" (or higher) permissions. As a repository owner I can grant these permissions only to individual users (which I have done for you) or teams (there is no "All Mozilla" team as far as I know). Only Github Admins can change base permissions that would apply to all repositories and all org members.

TIL.  I see in https://github.com/orgs/mozilla/teams that there's an https://github.com/orgs/mozilla/teams/admin-all-org-members team, but I'm not sure it's wise to use it for this purpose.  I will ask a GH admin what can be done here.

jetstream/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback.toml
danielkberry: It is determined automatically from Experimenter and doesn't need to be named anything in particular. It's done as a part of branch setup in Experimenter: the first branch is marked as the control branch so that's where we could put the `disabled` branch. The `enabled` branch could go next. In this case I think the branches in Experimenter got swapped, so that's why we need the manual override. 
Screenshot 2023-02-06 at 9 28 44 AM

danielkberry (approved):

PR #37, opened 2023-02-09
Add custom enrollment query

danielkberry:
Enrollment query will filter to data collected after the treatment has had time to be applied (ignore first 2 sessions after enrollment)


danielkberry:
@scholtzan it looks like the normal run for this experiment ran out of memory (see [this error](https://experimenter.services.mozilla.com/nimbus/abouthome-startup-cache-3/results) in Experiment). What's the best way to coax this run through?

scholtzan (approved):

PR #39, opened 2023-02-10
New profile activation definitions for Fenix and iOS

bani:
This PR adds the definitions for the [Mobile Activation Metric](https://docs.google.com/document/d/1ab2bRcmOdkxgLu-Xin7fZ25qFZGY5IyyMEkVCB1jSoQ/edit#), for future use in experiments.

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -381,6 +381,4 @@ submission_date_column = "submission_date"
 
 [data_sources.new_profile_activation]
 from_expression = "`moz-fx-data-shared-prod.fenix.new_profile_activation`"
-experiments_column_type = "simple"
-client_id_column = "client_id"
-submission_date_column = "submission_date"
+experiments_column_type = "none"
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -313,6 +313,13 @@ friendly_name = "Sponsored Tiles Preference Toggled"
 description = "Number of times Contile Sponsored Tiles setting is flipped."
 owner = "xluo-all@mozilla.com"
 
+[metrics.new_profile_activation]
+select_expression = "COALESCE(SUM(activated))"
+data_source = "new_profile_activation" 
+friendly_name = "New Profile Activation"
+description = "A new profile is counted as activated one week after creation if it meets the following conditions: 1) at least 3 days of use during first week 2) at least one search between days 4-7."
+owner = "vsabino@mozilla.com"
+
 
 [data_sources]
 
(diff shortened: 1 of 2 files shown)


definitions/fenix.toml
```diff
@@ -371,3 +378,9 @@ from_expression = """(
 experiments_column_type = "simple"
 client_id_column = "client_id"
 submission_date_column = "submission_date"
+
+[data_sources.new_profile_activation]
+from_expression = "`moz-fx-data-shared-prod.fenix.new_profile_activation`"
+experiments_column_type = "simple"
+client_id_column = "client_id"
+submission_date_column = "submission_date"
```
bani, line 386: What is the submission date used for? My table is partitioned on submission_date, but for most reporting it'd be ideal to use first_seen_date, which is lagged by 7 days.

danielkberry: Submission date is used for metrics in order to determine which records should be counted, mozanalysis [looks for rows ](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/metrics/__init__.py#L158-L162) where the submission date is within a particular observation window. So for example, for the week-2 results it looks for rows where the submission date is 7 to 14 days after the enrollment date. 

For your new activation metric, does submission date mean one of:
1. The date they activated (could be any time within first week after first seen)
2. Always 7 days after first seen

Option 2 will work as written, option 1 won't

definitions/fenix.toml
```diff
@@ -371,3 +378,9 @@ from_expression = """(
 experiments_column_type = "simple"
 client_id_column = "client_id"
 submission_date_column = "submission_date"
+
+[data_sources.new_profile_activation]
+from_expression = "`moz-fx-data-shared-prod.fenix.new_profile_activation`"
+experiments_column_type = "simple"
```
danielkberry, line 384: I think we might want `experiments_column_type = "none"` here. The [other options](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/metrics/__init__.py#L105-L128) are for tables enriched with experiment information and used primarily to filter out metrics from the day of enrollment. I don't think we need to worry about that here

definitions/firefox_ios.toml
```diff
@@ -225,3 +232,9 @@ from_expression = """(
 experiments_column_type = "simple"
 client_id_column = "client_id"
 submission_date_column = "submission_date"
+
+[data_sources.new_profile_activation]
+from_expression = "`moz-fx-data-shared-prod.firefox_ios.new_profile_activation`"
+experiments_column_type = "simple"
+client_id_column = "client_id"
```
danielkberry, line 239: `client_id_column` and `submission_date_column` are both [default values](https://github.com/mozilla/metric-config-parser/blob/main/metric_config_parser/data_source.py#L54-L55) so I think they could be left off

definitions/fenix.toml
bani: Ok, great. It's option 2, it's computed exactly 7 days after.

danielkberry (approved):
LGTM!

PR #49, opened 2023-02-16
Remove KDE statistics from `abouthome-startup-cache-3`

danielkberry:


jetstream/abouthome-startup-cache-3.toml
```diff
@@ -50,7 +50,34 @@ FROM (
...
+[data_sources.clients_daily]
+from_expression = """(
+        SELECT * FROM
+                mozdata.telemetry.clients_daily d
+        WHERE `mozfun.map.get_key`(
+                d.experiments, 'abouthome-startup-cache-3'
+        ) IS NOT NULL
+)"""
+friendly_name = "Clients Daily"
+description = "Clients Daily"
```
scholtzan, line 80: @danielkberry  This should be ok, right? The query did finish when filtering out clients based on the experiment slug before doing the join

danielkberry: This is a genius idea

PR #52, opened 2023-02-21
Calculate some default metrics at daily level

danielkberry:


ncalexan (approved):
Certainly!  Might as well put this in the 1pct TOML as well, so that folks who copy-paste don't see divergence.

danielkberry:
Going to wait to merge until issues with new `--analysis_bases` option are resolved

PR #58, opened 2023-03-06
[opmon] First definitions for Desktop Integrations operational monito…

ncalexan:
…ring.

This covers the first few operational metrics for:

- uninstalls
- background update
- the Windows Default Browser Agent

These definitions are intended to populate the various statistics tables and views.  `opmon` doesn't have the support we'd need for per-dimension alerts so we anticipate handling such alerting in another product (Grafana) or in the future (see
https://github.com/mozilla/opmon/issues/119).

### DIFF
--- a/opmon/firefox-background-update.toml
+++ b/opmon/firefox-background-update.toml
@@ -1,0 +1,160 @@
+# Monitoring of Firefox background update process.
+
+[project]
+
+name = "Firefox Background Update"
+platform = "firefox_desktop"
+xaxis = "submission_date"
+start_date = "2022-01-01"
+# We want continuous monitoring of this data:
+# end_date =
+skip_default_metrics = true
+
+metrics = [
+    "client_volume",
+    "ping_volume",
+    "success_rate",
+    "exception_rate",
+    # It turns out that the success rate is ~1.0 and the exception rate ~0.0, so
+    # the exception volume is more useful.
+    "exception_volume",
+    "crash_volume",
+    # TODO: "exit_state_rate",
+]
+
+[project.population]
+
+data_source = "background_update"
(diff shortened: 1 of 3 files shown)


scholtzan (approved):
Those look good to me

scholtzan:
You should be able to merge it whenever you want them to land

ncalexan:
> You should be able to merge the PR whenever you want them to land

Thanks, @scholtzan.  I've asked for a lot of historical data: will the pipeline figure out how to do this safely/efficiently?

scholtzan:
The backfill for historical data doesn't run automatically. I can trigger backfill back to 2022-01-01 once this has been merged.
Querying crashes is likely where most of the cost will come from so we could add some sampling. But when I checked the amount of data scanned looked relatively low, so I don't think that is strictly necessary.

ncalexan:
> The backfill for historical data doesn't run automatically. I can trigger backfill back to 2022-01-01 once this has been merged. Querying crashes is likely where most of the cost will come from so we could add some sampling. But when I checked the amount of data scanned looked relatively low, so I don't think that is strictly necessary.

OK!  When you're in the area, a few lines about how backfill and associated costs work would be helpful (in https://docs.telemetry.mozilla.org/cookbooks/operational_monitoring.html).  I assumed you'd stop bad things happening if this was going to be particularly expensive, but it was definitely a question I thought about while building these.

scholtzan:
We do have some mechanisms to prevent incredibly things from happening. There are some limits that any query running on BigQuery cannot exceed. 
But yeah, might be good for the preview command to show cost estimates.

scholtzan:
Backfill is done

PR #60, opened 2023-03-08
Skip experiment

scholtzan:
This experiment has never been launched, but jetstream is picking it up for analysis due to Experimenter exposing it as a real experiment.


mikewilli (approved):
This seems like a reasonable approach for now, and we can look at updating the API to avoid this kind of issue going forward.

PR #63, opened 2023-03-13
[skip ci] Fxa sign in

danielkberry:


jetstream/definitions/firefox_desktop.toml
```diff
@@ -80,4 +80,6 @@ pre_treatments = ["remove_nulls"]
 
 [metrics.imported_history.statistics.binomial]
 [metrics.imported_history_count.statistics.bootstrap_mean]
-pre_treatments = ["remove_nulls"]
\ No newline at end of file
+pre_treatments = ["remove_nulls"]
+
+[metrics.fxa_signed_in.statistics.binomial]
```
scholtzan, line 85: Do we want experiments that are currently live to get rerun?

scholtzan (approved):

jetstream/definitions/firefox_desktop.toml
danielkberry: No, we don't. I changed the title on the PR, is that the right way to configure it to skip the CI?

scholtzan: Yeah. I'll also update the table timestamps, otherwise the next time a rerun is triggered the experiments will see that the config got updated after their tables were populated and do a rerun

PR #66, opened 2023-03-16
add Desktop DAU and DAU 28MA

bochocki:
Adds definitions for desktop DAU and DAU 28MA.

I added more context around metric ownership in the description because we want to have clear signposts for issues with these metrics, and it seems like the `owner` field is not rendered on [mozilla.github.io](https://mozilla.github.io/metric-hub). Please let me know if there's a better way to make ownership/contact information available.

Ticket: https://mozilla-hub.atlassian.net/browse/DS-2680

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -267,7 +267,7 @@ description = """
     The average number of unique clients with >0 active hours and >0 URIs loaded that
     we received a main ping from each day in the 28-day period ending on December 15th. 
     To reconstruct the annual Desktop DAU KPI, this metric needs to be aggregated by
-    `EXTRACT(YEAR FROM submission_date)`.
+    `EXTRACT(YEAR FROM submission_date)`.  
 
     For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
 """
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -239,48 +239,40 @@ description = """
 
 [metrics.daily_active_users]
 data_source = "clients_daily"
-select_expression = """COUNTIF(
-    active_hours_sum > 0 AND
-    COALESCE(
-        scalar_parent_browser_engagement_total_uri_count_normal_and_private_mode_sum,
-        scalar_parent_browser_engagement_total_uri_count_sum,
-        0
-    ) > 0
-)"""
+select_expression = "COUNTIF(active_hours_sum > 0 AND total_uri_count > 0)"
 type = "scalar"
 friendly_name = "DAU"
 description = """
-    The number of unique clients that we received a main ping from on each `submission_date`,
-    given that the client had >0 active hours and >0 URIs loaded.
+    The number of unique clients with >0 active hours and >0 URIs loaded that
+    we received a main ping from each day. To be comparable to DAU used for KPI
+    tracking, this metric needs to be aggregated by `submission_date`.  
 
-    For questions, please contact bochocki@mozilla.com or the Browser Data Science team.
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
 """
 category = "KPI"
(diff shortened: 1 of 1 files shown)


definitions/firefox_desktop.toml
```diff
@@ -237,6 +237,53 @@ description = """
...
+
+[metrics.daily_active_users_28_day_moving_average]
+data_source = "clients_daily"
+select_expression = """AVG(COUNTIF(
+    active_hours_sum > 0 AND
+    COALESCE(
+        scalar_parent_browser_engagement_total_uri_count_normal_and_private_mode_sum,
+        scalar_parent_browser_engagement_total_uri_count_sum,
+        0
+    ) > 0
```
jdma-mozilla, line 271: iirc, there is a new probe (backfilled) that combines these uri count probes. shall we use the new one instead of the old approach? @irrationalagent should be able to provide more context.

jdma-mozilla (changes requested):

danielkberry:
@bochocki I'm not sure entirely the right way to handle this, but are these metrics intended to come out at the day level? Right now, metric definitions can be aggregated over arbitrary periods. In experiments we often use `week` and `overall` levels, so we'd sum up "DAU" by client over a week period. So in that situation, the metric would be basically DOU, right? Is that an issue?

bochocki:
> @bochocki I'm not sure entirely the right way to handle this, but are these metrics intended to come out at the day level? Right now, metric definitions can be aggregated over arbitrary periods. In experiments we often use `week` and `overall` levels, so we'd sum up "DAU" by client over a week period. So in that situation, the metric would be basically DOU, right? Is that an issue?

That's a good point @danielkberry, this definition is only valid at the day level. When this metric is aggregated it isn't DAU anymore, but it doesn't become WAU or MAU either -- it's "cumulative DAU" which, as you said, is DOU. I think that for that reason we will want to avoid simply summing this metric over time, including when running experiments.

Is it possible to set some default aggregation to `average` instead of a `sum`? That way the metric becomes the "average DAU over the period", and in a 28-day period would correspond exactly to the DAU 28MA KPI. These aggregations will always be on the same scale as DAU (and DAU 28MA), so understanding magnitude changes in aggregations of the metric would be intuitive.

danielkberry (approved):
Following the discussion as a team, these metrics are good as-is & we'll implement the changes I suggested through Jetstream/mozanalysis instead.

bochocki:
We talked about this offline and decided to take the following steps:
1. Include a precise definition of the Desktop DAU KPI (`metrics.desktop_dau_kpi`). This metric is intended to be used more for reference than for regular use.
2. Include a definition of DAU (`metrics.daily_active_users`) that can be used in dashboards or queries if needed. This metric is *not* invariant across submission dates or client segmentation. Any logic related to re-scaling of the metric will need to be implemented by the user of the metric.

definitions/firefox_desktop.toml
```diff
@@ -237,6 +237,45 @@ description = """
 """
 
 
+[metrics.daily_active_users]
+data_source = "clients_daily"
+select_expression = "COUNTIF(active_hours_sum > 0 AND total_uri_count > 0)"
```
irrationalagent, line 242: Do we need to add to note saying that if you DON'T aggregate by date, you are getting days of use ("or active user-days", or something?) Or does that just confuse things?I ask because this is really only "DAU" in one particular configuration - in all other cases you are opening up the possibility that a client counts more than once. 

I'm not blocking on this, but my 2c.

irrationalagent (commented):
I'm OK with this. I guess we decided we don't need an additional "average dau" that divides by distinct submission date?

definitions/firefox_desktop.toml
bochocki: Sounds good to me! I added "If the metric is NOT aggregated by `submission_date`, the metric is similar to a "days of use" metric."

PR #68, opened 2023-03-21
Focus legacy to glean conversion

pissac17:
The field "search_count_imputed"  converts legacy search count data prior to 2023-01-01 to glean using legacy to glean multiplier  of 0.49. This field is introduced  because we observed that the search counts for glean and legacy differ  by big margin for Focus android . Hence we developed a multiplier for converting historical search counts from legacy to glean for focus android. For more details please refer to : https://docs.google.com/document/d/1mx9bdm2jXlcO4UsqdzaSvOE4Ig2ZoDKeDPMR1XiRMvI/edit#

### DIFF
--- a/definitions/focus_android.toml
+++ b/definitions/focus_android.toml
@@ -75,7 +75,7 @@ owner = "xluo-all@mozilla.com"
 [metrics.search_count_imputed]
 friendly_name = "SAP search count (legacy to glean)"
 description = "Imputed SAP for converting historical search counts from legacy to glean"
-select_expression = "{{agg_sum('search_count') * 0.45}}"
+select_expression = "SUM(IF(submission_date <= '2023-01-01', 'search_count'* 0.45,'search_count'))"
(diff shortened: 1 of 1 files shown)


definitions/focus_android.toml
```diff
@@ -72,6 +72,14 @@ category = "search"
 type = "scalar" 
 owner = "xluo-all@mozilla.com"
 
+[metrics.search_count_imputed]
+friendly_name = "SAP search count (legacy to glean conversion)"
+description = "Imputed SAP for converting historical search counts from legacy to glean"
+select_expression = "ROUND(SUM(IF(submission_date < '2023-01-01', 'search_count'* 0.49,'search_count')))"
```
scholtzan, line 78:
```suggestion
select_expression = "ROUND(SUM(IF(submission_date < '2023-01-01', search_count * 0.49, search_count)))"
```

scholtzan (approved):
Looks good implementation wise. Curtis should confirm that the metric definition itself is correct.

PR #78, opened 2023-03-27
Query optimizations

danielkberry:


danielkberry:
@florezita The events & main ping tables are very large and can sometimes cause issues. These filter the tables to only data from your experiment, which makes Jetstream run faster.

PR #82, opened 2023-04-04
Rs650 add focus i os search offset to metrics hub

pissac17:
in Focus iOS, search numbers reported by Glean do not match the ones reported by legacy. After investigating, we found that that is due to a [product bug](https://github.com/mozilla-mobile/focus-ios/issues/3703) that makes it so direct URLs typed in the address bar are counted as search. This inflates Glean reported search numbers in Focus iOS by roughly 40%.
search_count_imputed field  accounts for this discrepancy in metrics-hub

### DIFF
--- a/definitions/focus_ios.toml
+++ b/definitions/focus_ios.toml
@@ -76,7 +76,7 @@ owner = "xluo-all@mozilla.com"
 [metrics.search_count_imputed]
 friendly_name = "SAP search count (legacy to glean conversion)"
 description = "Imputed SAP for converting historical search counts from legacy to glean"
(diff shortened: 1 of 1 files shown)


definitions/focus_ios.toml
```diff
@@ -73,6 +73,15 @@ category = "search"
 type = "scalar"
 owner = "xluo-all@mozilla.com"
 
+[metrics.search_count_imputed]
```
Dexterp37, line 76:
```suggestion
# To get more context about how the conversion factors were derived, see
# https://mozilla-hub.atlassian.net/browse/RS-650
[metrics.search_count_imputed]
```

Dexterp37 (approved):
R+ with a comment just to make it easier for future folks to trace it back to where the investigation happaned and collect more context.

PR #86, opened 2023-04-18
add mobile DAU definitions

bochocki:
Ticket: [DS-2680](https://mozilla-hub.atlassian.net/browse/DS-2680)

This adds product-specific DAU metrics for the four mobile products that contribute to our Mobile KPI: Fenix, Firefox iOS, Focus Android, and Focus iOS.

Note that the retention period on `focus_android.baseline` is currently only 180 days, but [there's an open ticket to expand that to 760 days](https://bugzilla.mozilla.org/show_bug.cgi?id=1828666). When calculating DAU for reporting, we've been using `focus_android.baseline_clients_daily` instead of `focus_android.baseline`:

```sql
SELECT submission_date,
       COUNT(1) AS dau
  FROM focus_android.baseline_clients_daily
 WHERE submission_date >= start_date
   AND (isp != 'BrowserStack' OR isp IS NULL)
 GROUP BY 1
```

But for the purpose of experiments -- which Metric Hub is most commonly used for -- the `baseline` table includes useful columns (like `ping_info.experiments.*`) that are not available in the `baseline_clients_daily` table. So I opted to use `focus_android.baseline` here. This will keep its DAU behavior consistent with other mobile products on Metric Hub.

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -357,6 +357,20 @@ friendly_name = "Baseline"
 description = "Baseline Ping"
 build_id_column = "REPLACE(CAST(DATE(mozfun.norm.fenix_build_to_datetime(client_info.app_build)) AS STRING), '-', '')"
 
+[data_sources.baseline_v2]
+from_expression = """(
+    SELECT
+        p.*,
+        DATE(p.submission_timestamp) AS submission_date
+    FROM `moz-fx-data-shared-prod.{dataset}.baseline` p
+)"""
+client_id_column = "client_info.client_id"
+experiments_column_type = "glean"
+default_dataset = "fenix"
+friendly_name = "Baseline"
+description = "Baseline Ping"
+build_id_column = "REPLACE(CAST(DATE(mozfun.norm.fenix_build_to_datetime(client_info.app_build)) AS STRING), '-', '')"
+
 [data_sources.events]
 from_expression = """(
     SELECT
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -28,7 +28,8 @@ description = """
     The number of unique clients that we received a baseline ping from each day, excluding
     pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
     this metric needs to be aggregated by `submission_date`. If the metric is NOT
-    aggregated by `submission_date`, the metric is similar to a "days of use" metric.
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric. For more details, refer to
+    [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
 
     For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
 """
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -20,7 +20,7 @@ data_source = "baseline"
 
 
 [metrics.daily_active_users]
-data_source = "baseline"
+data_source = "baseline_v2"
 select_expression = "COUNT(DISTINCT CASE WHEN LOWER(metadata.isp.name) != 'browserstack' THEN client_info.client_id ELSE NULL END)"
 type = "scalar"
 friendly_name = "DAU"
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -18,6 +18,25 @@ description = "The number of days in an observation window that clients used the
 select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
 data_source = "baseline"
 
+
+[metrics.daily_active_users]
+data_source = "baseline"
+select_expression = "COUNT(DISTINCT CASE WHEN LOWER(metadata.isp.name) != 'browserstack' THEN client_info.client_id ELSE NULL END)"
+type = "scalar"
+friendly_name = "DAU"
+description = """
+    The number of unique clients that we received a baseline ping from each day, excluding
+    pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric.
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
+deprecated = false
+
+
 [metrics.user_reports_site_issue_count]
 data_source = "events"
 select_expression = """
--- a/definitions/firefox_ios.toml
+++ b/definitions/firefox_ios.toml
@@ -33,7 +33,7 @@ data_source = "baseline"
 
 
 [metrics.daily_active_users]
-data_source = "baseline"
+data_source = "baseline_v2"
 select_expression = "COUNT(DISTINCT CASE WHEN LOWER(metadata.isp.name) != 'browserstack' THEN client_info.client_id ELSE NULL END)"
 type = "scalar"
 friendly_name = "DAU"
--- a/definitions/firefox_ios.toml
+++ b/definitions/firefox_ios.toml
@@ -31,6 +31,25 @@ description = "The number of days in an observation window that clients used the
 select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
 data_source = "baseline"
 
+
+[metrics.daily_active_users]
+data_source = "baseline"
+select_expression = "COUNT(DISTINCT CASE WHEN LOWER(metadata.isp.name) != 'browserstack' THEN client_info.client_id ELSE NULL END)"
+type = "scalar"
+friendly_name = "DAU"
+description = """
+    The number of unique clients that we received a baseline ping from each day, excluding
+    pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric.
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
+deprecated = false
+
+
 ## search metrics
 
 [metrics.organic_search_count]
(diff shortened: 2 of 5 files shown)


definitions/fenix.toml
```diff
@@ -18,6 +18,26 @@ description = "The number of days in an observation window that clients used the
 select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
 data_source = "baseline"
 
+
+[metrics.daily_active_users]
+data_source = "baseline"
```
irrationalagent, line 23: This datasource seems to point at [moz-fx-data-shared-prod.org_mozilla_firefox.baseline](https://github.com/mozilla/metric-hub/blob/main/definitions/fenix.toml#L331) which IIUC is just the fenix release channel baseline data. In [past queries](https://sql.telemetry.mozilla.org/queries/91619) we have pointed at `fenix.baseline` which [contains data from all channels](https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,mozdata.fenix.baseline,PROD)/Lineage?is_lineage_mode=true&). I believe the intention is to count clients from all release channels... the non-release counts are going to be very small but I think if we want things to align with the KPI we would need to count those channels.

We should get a data engineer to confirm my claim about `org_mozilla_firefox.baseline` only reflecting the release channel. If its true, we should ascertain if there is a reason why metric hub is using this as its source for baseline data, rather than a source that contains all channels.  

(Similarly, I think the active users aggregates tables are based on the fenix last_seen_joined tables, which would also include the non-release channels. Again we should confirm with @lucia-vargas-a or someone else since there has been recent work there).

definitions/firefox_ios.toml
```diff
@@ -31,6 +31,26 @@ description = "The number of days in an observation window that clients used the
 select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
 data_source = "baseline"
 
+
+[metrics.daily_active_users]
+data_source = "baseline"
```
irrationalagent, line 36: Similar comment to above, I believe we would be missing any baseline data from ios beta channel here AFAICT that data lives in the ` moz-fx-data-shared-prod.org_mozilla_ios_firefoxbeta` associated tables. The all-channel unioned data would be in `moz-fx-data-shared-prod.firefox_ios` (again, IIUC).

irrationalagent (changes requested):

irrationalagent:
We should confirm my claims above about the metric hub data sources only containing release channel data. If its true, we should ascertain why we are using those. This would have other consequences as well, for example if we ever tried to run an experiment on a non-release mobile channel then jetstream wouldn't be able to pull any metrics from these baseline tables.

definitions/fenix.toml
bochocki: That's a great catch; I misunderstood the dataset syntax. If I understand you correctly, it sounds like we might want to change the `[data_sources.baseline]` definition in each `toml` file from:

```yaml
default_dataset = "org_mozilla_firefox"
```

to this:
```yaml
default_dataset = "fenix"  # or whatever the corresponding product is
```

danielkberry: > I believe the intention is to count clients from all channels... the non-release counts are going to be very small but I think if we want things to align with the KPI we would need to count those channels.
I'm surprised the KPI includes non-release clients, I would've thought it would be release-only. 

Leif is right, `moz-fx-data-shared-prod.org_mozilla_firefox` is Fenix release only. `mozdata.fenix` unions data from all channels.

irrationalagent: yea that's the idea. but I wonder if there is some reason that I don't know about, e.g. for jetstream, that we use  datasets from `org_mozilla_firefox` instead of `fenix`. @danielkberry or @scholtzan any idea? would it break things to make this switch? if not, should we open a separate PR to do that bit first before adding these?

danielkberry: So jetstream/opmon fill in the dataset depending on the app the experiment is running on. So a fenix release experiment gets `org_mozilla_firefox` while a beta experiment gets `org_mozilla_firefox_beta`. The _default dataset_ would be release, but if analysis was needed for another channel, it would switch to that channel

danielkberry: The dataset isn't hardcoded in the from expression, it's a variable that's filled in at runtime

irrationalagent: > I'm surprised the KPI includes non-release clients, I would've thought it would be release-only.

Our precedent is to count clients regardless of channel - the unified_metrics table for example queries from the views that union all the channels together (for mobile). [And for desktop there's no filter for the release channel specifically](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L259-L278). 

 Maybe its debatable whether we SHOULD count pre-release, but it seems to be the precedent anyway.

irrationalagent: @bochocki I wonder if its possible to pass a `dataset = fenix` line into the metric definition so we don't have to change the default

bochocki: So do we need to make changes here? Or leave everything as-is and rely on end-users to specify the correct dataset?

danielkberry: Maybe the baseline dataset should be modified to pull from `fenix` instead, since that's what the KPI is based on.

bochocki: Following @danielkberry's comment, maybe it makes sense to change the dataset since in most of the cases where different datasets are used (opmon/jetstream) the dataset will be specified explicitly, but in other cases (somebody copy-pasting the metric definition) it might not be obvious that something other than the default dataset should be used.

danielkberry: IMO, we should change metrics to align with the KPI, so if the KPI is using the `fenix` dataset, then as much as possible, our metrics should too. If we change the definition of the `baseline` dataset to point to `fenix`, then we won't have a notion of default dataset, we'll just pull directly from that one. Which I think is fine.

scholtzan: > We should get a data engineer to confirm my claim about org_mozilla_firefox.baseline only reflecting the release channel

I can confirm that this dataset only contains release. 


The reason why it isn't `fenix` to begin with is that when the `baseline` dataset was defined `fenix` didn't exist. Back in the days we didn't have these unified datasets that combine all channels, instead we had to explicitly specify the channel-dataset. Unfortunately, that notion is still encoded in jetstream and mozanalysis, so simply changing the definition of `baseline` could potentially cause some problems in jetstream (we'd need to make some changes there first). 

Also, to avoid breaking existing things, we probably shouldn't simply change the `baseline` definition. Instead, I'd recommend to add a `baseline_v2` data source with the new definition.

bochocki: Thanks @scholtzan, that makes a lot of sense. If it sounds good to everyone, I'll add `baseline_v2` datasets in each of the `toml` files that explicitly point to the relevant `<app_name>.baseline` tables.

irrationalagent (approved):
r+

definitions/fenix.toml
```diff
@@ -18,6 +18,26 @@ description = "The number of days in an observation window that clients used the
...
+    The number of unique clients that we received a baseline ping from each day, excluding
+    pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric. For more details, refer to
+    [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
```
scholtzan, line 37:
```suggestion
owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
```

definitions/firefox_ios.toml
```diff
@@ -31,6 +31,26 @@ description = "The number of days in an observation window that clients used the
...
+    The number of unique clients that we received a baseline ping from each day, excluding
+    pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric. For more details, refer to
+    [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
```
scholtzan, line 50:
```suggestion
owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
```

definitions/focus_android.toml
```diff
@@ -37,6 +37,27 @@ select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
...
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric. Note
+    that data retention on `focus_android.baseline` has [historically been only 180 days](https://bugzilla.mozilla.org/show_bug.cgi?id=1828666),
+    which is shorter than other mobile `baseline` tables. For more details, refer to
+    [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
```
scholtzan, line 57:
```suggestion
owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
```

definitions/focus_ios.toml
```diff
@@ -36,6 +36,25 @@ select_expression = "COUNT(DISTINCT DATE(submission_timestamp))"
...
+    The number of unique clients that we received a baseline ping from each day, excluding
+    pings originating from BrowserStack. To be comparable to DAU used for KPI tracking,
+    this metric needs to be aggregated by `submission_date`. If the metric is NOT
+    aggregated by `submission_date`, the metric is similar to a "days of use" metric. For more details, refer to
+    [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+
+    For questions, please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = "bochocki@mozilla.com, firefox-kpi@mozilla.com"
```
scholtzan, line 54:
```suggestion
owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
```

scholtzan (approved):
lgtm, `owner` can be a list of emails

bochocki:
Merging without additional review because the only commits included minor improvements suggested by Anna.

PR #90, opened 2023-04-21
add `submission_date_column` and `client_id_column`

bochocki:
- Add `submission_date_column` to `baseline_v2` data sources in for KPI-relevant mobile products.
- Add `submission_date_column` and `client_id_column` to `data_sources.clients_daily` for `firefox_desktop`.

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -1233,6 +1233,8 @@ description = "Search Clients Engines Sources Daily"
 
 [data_sources.clients_daily]
 from_expression = "mozdata.telemetry.clients_daily"
+submission_date_column = "submission_date"
+client_id_column = "client_id"
 friendly_name = "Clients Daily"
 description = "Clients Daily"
 
--- a/definitions/firefox_ios.toml
+++ b/definitions/firefox_ios.toml
@@ -219,6 +219,7 @@ from_expression = """(
     FROM `moz-fx-data-shared-prod.{dataset}.baseline` p
 )"""
 client_id_column = "client_info.client_id"
+submission_date_column = "DATE(submission_timestamp)"
 experiments_column_type = "glean"
 default_dataset = "firefox_ios"
 friendly_name = "Baseline"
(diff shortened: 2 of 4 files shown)


definitions/firefox_ios.toml
```diff
@@ -219,6 +219,7 @@ from_expression = """(
     FROM `moz-fx-data-shared-prod.{dataset}.baseline` p
 )"""
 client_id_column = "client_info.client_id"
+submission_date_column = "DATE(submission_timestamp)"
```
danielkberry, line 222: I've never seen an expression here that wasn't just the label of a column name. Does this work?

bochocki: I was going off of the `build_id_column` pattern, which is often a large-ish SQL snippet, e.g.:

```SQL
"REPLACE(CAST(DATE(mozfun.norm.fenix_build_to_datetime(client_info.app_build)) AS STRING), '-', '')"
```
But I think that a SQL snippet will work here, too. [From DTMO](https://docs.telemetry.mozilla.org/concepts/metric_hub.html#data_sources-section):

```
Data sources specify the tables data should be queried from.

[data_sources]

[data_sources.main_v1]
# FROM expression - often just a fully-qualified table name. Sometimes a subquery.
from_expression = "mozdata.telemetry.main"

# SQL snippet specifying the submission_date column
```

definitions/firefox_desktop.toml
```diff
@@ -1233,6 +1233,8 @@ description = "Search Clients Engines Sources Daily"
 
 [data_sources.clients_daily]
 from_expression = "mozdata.telemetry.clients_daily"
+submission_date_column = "submission_date"
+client_id_column = "client_id"
```
mikewilli, line 1237: These are both the default values so I don't think they're strictly necessary, but that said I have no problem being explicit about it.

definitions/firefox_ios.toml
mikewilli: I don't see any problems with this. The `submission_date_column` is used in the select (e.g., something like `SELECT {{ datasource.submission_date_column }} AS submission_date`) and then the `submission_date` alias is used elsewhere in [the template](https://github.com/mozilla/metric-config-parser/blob/main/metric_config_parser/templates/metrics_query.sql).

mikewilli (approved):

PR #99, opened 2023-05-02
[ci rerun-skip] Rerun skip support

scholtzan:
for https://github.com/mozilla/jetstream/issues/1627

depends on https://github.com/mozilla/jetstream/pull/1714


mikewilli (approved):
Looks good to me! I was going to comment on using `ubuntu-latest` in case that auto-updates out from underneath us, but I noticed that we're already using it for other actions, so I think we can tackle that separately if it becomes a problem.

PR #105, opened 2023-05-11
Create iOS Onboarding Outcome

danielkberry:


danielkberry:
Here's a start of an Onboarding outcome. We can add more metrics as we go along. 

@bani @rhythmofrain63 : are these metrics still correct? I just copied them from the default browser Outcome

bani (approved):

PR #109, opened 2023-05-22
add and update DAU definitions

bochocki:
Makes several changes to the DAU definitions to simplify definitions and improve discoverability for the five products that contribute towards DAU: Firefox Desktop, Fenix, Firefox iOS, Focus Android, and Focus iOS.

Changes
- Added a `daily_active_users` and `client_level_daily_active_users` metric for each of the five relevant products.
  - `daily_active_users` is simply the sum of DAU in the `active_users_aggregates` table; this is the version of DAU that should generally be used for reporting.
  - `client_level_daily_active_users` is a version of DAU that provides client-level information. This is not used for reporting and is not guaranteed to be exactly equal to `daily_active_users` for any product on a particular date.
- Added `data_sources.active_users_aggregates` for each product; the data source is filtered by the relevant product to improve query efficiency.
- Added `cross_product.toml`, a place to store metrics that need to be applied across products. The use case here is for mobile DAU, which needs to be calculated across 4 mobile products. I added the relevant definitions to that file.
- Deprecated previous versions of DAU definitions when relevant.

### DIFF
--- a/definitions/cross_product.toml
+++ b/definitions/cross_product.toml
@@ -1,6 +1,6 @@
 [metrics]
 
-[metrics.mobile_daily_active_users]
+[metrics.mobile_daily_active_users_v1]
 data_source = "mobile_active_users_aggregates"
 select_expression = "SUM(dau)"
 type = "scalar"
--- a/definitions/cross_product.toml
+++ b/definitions/cross_product.toml
@@ -1,0 +1,53 @@
+[metrics]
+
+[metrics.mobile_daily_active_users]
+data_source = "mobile_active_users_aggregates"
+select_expression = "SUM(dau)"
+type = "scalar"
+friendly_name = "Mobile DAU"
+description = """
+    This is the official DAU reporting definition. The logic is
+    [defined in `bigquery-etl`](https://github.com/mozilla/bigquery-etl/tree/main/sql_generators/active_users/templates)
+    and is automatically cross-checked, actively monitored, and change controlled.
+    Whenever possible, this is the preferred DAU reporting definition to use for Mobile products.
+    This metric needs to be aggregated by `submission_date`. If it is not aggregated by `submission_date`,
+    it is similar to a "days of use" metric, and not DAU.
+    
+    For more information, refer to [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+    For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
+deprecated = false
+
+[metrics.mobile_dau_kpi]
+data_source = "mobile_active_users_aggregates"
+select_expression = "SUM(IF(FORMAT_DATE('%m-%d', submission_date) BETWEEN '11-18' AND '12-15', dau, 0)) / 28"
+type = "scalar"
+friendly_name = "Mobile DAU KPI"
+description = """
+    The average [Mobile DAU](https://mozilla.github.io/metric-hub/metrics/cross_product/#mobile_daily_active_users) in the 28-day period ending on December 15th. This is the official
+    Mobile DAU KPI reporting definition. The logic for calculating DAU is
+    [defined in `bigquery-etl`](https://github.com/mozilla/bigquery-etl/tree/main/sql_generators/active_users/templates)
+    and is automatically cross-checked, actively monitored, and change controlled.
+    To reconstruct the annual Mobile DAU KPI, this metric needs to be aggregated by
+    `EXTRACT(YEAR FROM submission_date)`.  
+
+    For more information, refer to [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+    For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
+deprecated = false
+
+
+[data_sources]
+
+[data_sources.mobile_active_users_aggregates]
+from_expression = """(
+    SELECT *
+     FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates`
+    WHERE app_name IN ('Fenix',  'Firefox iOS', 'Focus Android', 'Focus iOS')
+)"""
+friendly_name = "Active Users Aggregates"
+description = "Active Users Aggregates, filtered on the Mobile product group"
+submission_date_column = "submission_date"
--- a/definitions/cross_product.toml
+++ b/definitions/cross_product.toml
@@ -1,7 +1,7 @@
 [metrics]
 
 [metrics.mobile_daily_active_users_v1]
-data_source = "mobile_active_users_aggregates"
+data_source = "mobile_active_users_aggregates_v1"
 select_expression = "SUM(dau)"
 type = "scalar"
 friendly_name = "Mobile DAU"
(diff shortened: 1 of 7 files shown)


definitions/firefox_ios.toml
```diff
@@ -197,6 +235,16 @@ owner = "vsabino@mozilla.com"
 
 [data_sources]
 
+[data_sources.active_users_aggregates]
+from_expression = """(
+    SELECT *, NULL as client_id
```
bochocki, line 240: Adding the `NULL as client_id` here is a hack; I'm not sure if there's a better way to do this. We want to use `active_users_aggregates` for these metrics because we have bigquery-etl infrastructure that ensures the values in this table are correct. However, this table doesn't include `client_id`, and we don't need it for calculating DAU.

definitions/cross_product.toml
```diff
@@ -0,0 +1,53 @@
+[metrics]
```
bochocki, line 1: Naming is hard; happy to rename or consider alternatives for this filename if there are any strong opinions out there.

definitions/cross_product.toml
```diff
@@ -0,0 +1,53 @@
...
+description = """
+    The average [Mobile DAU](https://mozilla.github.io/metric-hub/metrics/cross_product/#mobile_daily_active_users) in the 28-day period ending on December 15th. This is the official
+    Mobile DAU KPI reporting definition. The logic for calculating DAU is
+    [defined in `bigquery-etl`](https://github.com/mozilla/bigquery-etl/tree/main/sql_generators/active_users/templates)
+    and is automatically cross-checked, actively monitored, and change controlled.
+    To reconstruct the annual Mobile DAU KPI, this metric needs to be aggregated by
+    `EXTRACT(YEAR FROM submission_date)`.  
+
+    For more information, refer to [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+    For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
```
bochocki, line 36: I'm currently adding this contact information in the description, but it could be nice if the `owner` field populated in the website.

definitions/firefox_ios.toml
```diff
@@ -197,6 +235,16 @@ owner = "vsabino@mozilla.com"
...
+     FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates`
+    WHERE app_name = 'Firefox iOS'
+)"""
+friendly_name = "Active Users Aggregates"
+description = "Active Users Aggregates, filtered on Firefox iOS"
+submission_date_column = "submission_date"
```
scholtzan, line 246: you can also set `client_id_column` to NULL alternatively
```suggestion
from_expression = """(
    SELECT *
     FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates`
    WHERE app_name = 'Firefox iOS'
)"""
friendly_name = "Active Users Aggregates"
description = "Active Users Aggregates, filtered on Firefox iOS"
submission_date_column = "submission_date"
client_id_column = "NULL"  # this table doesn't include client_id, and we don't need it for calculating DAU
```

definitions/cross_product.toml
```diff
@@ -0,0 +1,53 @@
+[metrics]
+
+[metrics.mobile_daily_active_users]
```
scholtzan, line 3: For newly added metrics it is recommended to version them. Just to make it easier to deal with future changes

```suggestion
[metrics.mobile_daily_active_users_v1]
```

definitions/cross_product.toml
```diff
@@ -0,0 +1,53 @@
...
+    This metric needs to be aggregated by `submission_date`. If it is not aggregated by `submission_date`,
+    it is similar to a "days of use" metric, and not DAU.
+    
+    For more information, refer to [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
+    For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
+deprecated = false
+
+[metrics.mobile_dau_kpi]
```
scholtzan, line 22:
```suggestion
[metrics.mobile_dau_kpi_v1]
```

definitions/cross_product.toml
```diff
@@ -0,0 +1,53 @@
...
+    For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
+"""
+category = "KPI"
+owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
+deprecated = false
+
+
+[data_sources]
+
+[data_sources.mobile_active_users_aggregates]
```
scholtzan, line 45:
```suggestion
[data_sources.mobile_active_users_aggregates_v1]
```

definitions/firefox_ios.toml
bochocki: That's great to know, thank you!

definitions/cross_product.toml
mikewilli: `multi_product` was my thought as well, and I think that'd be my vote, but don't feel strongly about it.

definitions/cross_product.toml
mikewilli, line 1: Should the metrics defined here indicate which products they apply to, or is the `data_source` definition explicit enough to provide this info?

definitions/firefox_desktop.toml
```diff
@@ -1183,7 +1204,18 @@ description = "Normalized Country Code"
...
+[data_sources.active_users_aggregates_v1]
+from_expression = """(
+    SELECT *
+     FROM `moz-fx-data-shared-prod.telemetry.active_users_aggregates`
+    WHERE app_name = 'Firefox Desktop'
+)"""
+friendly_name = "Active Users Aggregates"
+description = "Active Users Aggregates, filtered on Firefox Desktop"
+submission_date_column = "submission_date"
+client_id_column = "NULL"  # this table doesn't include client_id, and we don't need it for calculating DAU
```
mikewilli, line 1217: @scholtzan is `"NULL"` here the correct value?

definitions/cross_product.toml
bochocki: I like `multi_product` and hadn't considered it. I'll make that change.

mikewilli (commented):
Looks good to me, main question was to confirm how to specify a null client_id.

definitions/cross_product.toml
bochocki: I can add the products to the description, but they're defined in the `data_source`. Or are you asking if we should add a field? E.g.:
```
products = ['Fenix', 'Firefox iOS', 'Focus Android', 'Focus iOS']
```

definitions/firefox_desktop.toml
bochocki: I _think_ this should work, based on Anna's previous comment: https://github.com/mozilla/metric-hub/pull/109#discussion_r1202645805

definitions/cross_product.toml
mikewilli: So I think between your response and some further digging I've answered this for myself. What you have now is perfectly fine, and adding to the description is up to you, but probably not necessary.

I was originally thinking we might want or need a field in the metric definition, but as you said, that the info is in the `data_source`, so this question was more about the ability to see this info readily in the docs. That said, even that is probably not necessary because I realized that the metric docs already have a link to the `data_source`.

definitions/firefox_desktop.toml
mikewilli: Sorry I thought I edited this comment before sending, but this didn't need to be exclusively a question for Anna. The point was just to confirm that this is the correct format for specifying a null client_id.

mikewilli: Ok, sounds good, thanks!

irrationalagent (approved):

mikewilli (approved):

PR #114, opened 2023-05-26
Added Networking outcome and starter metrics

acreskeyMoz:
The Network/Security/Privacy teams would like to create a shared outcome tracking network layer metrics including Necko, TLS, and NSS.

I've created this outcome as a starting point.


scholtzan (approved):
The way the outcome is defined is correct. The metric definitions also look good to me

acreskeyMoz:
Thank you!

PR #123, opened 2023-06-07
Update import-bookmarks-copy.toml

danielkberry:


jetstream/import-bookmarks-copy.toml
```diff
@@ -3,20 +3,61 @@ weekly = ['imported_bookmarks', 'imported_logins', 'imported_history', 'is_pinne
...
+data_source = 'clients_last_seen_with_activity'
 window_start = 0
 window_end = 0
 
+[segments.infrequent_users]
+select_expression = """COALESCE(ANY_VALUE(activity_segment_importbookmarks) = 'infrequent', FALSE)"""
+data_source = 'clients_last_seen_with_activity'
+window_start = 0
+window_end = 0
+
```
scholtzan, line 37: Do we want an `other` segment here too?

PR #125, opened 2023-06-09
Create limited-font-visibility-for-fingerprinters.toml

danielkberry:


danielkberry:
I tested these with a local rerun. It's kinda working in the sense that it's not crashing, but two of the metrics in the outcome have issues. I'm going to push these changes to trigger a run so we can get guardrail metrics. If we want to continue investigating the outcome, we can do so next week.

PR #136, opened 2023-06-20
Update stricter_pbm.toml

danielkberry:


danielkberry:
@scholtzan Just to confirm the new behavior: live experiments will use the old definition of this outcome, unless manually triggered? To get a completed experiment to use this new outcome, can I simply rerun the analysis for that experiment?

scholtzan:
You'd either need to add the `--recreate-enrollments` option or run `jetstream rerun-skip --experiment_slug=<slug>` first before doing `rerun`

danielkberry:
ohh, hmm. That's unfortunate, as I didn't do those.

scholtzan:
ah wait, `rerun` alone should be enough

PR #139, opened 2023-06-22
Update limited-font-visibility-for-fingerprinters.toml

danielkberry:


danielkberry:
This adds an exposure signal mirroring the first experiment

PR #147, opened 2023-07-05
add fenix fxa metric

irrationalagent:
this adds a boolean outcome for whether a fenix user is signed in to FxA

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -49,7 +49,7 @@ description = """
     Whenever possible, this is the preferred DAU reporting definition to use for Fenix.
     This metric needs to be aggregated by `submission_date`. If it is not aggregated by `submission_date`,
     it is similar to a "days of use" metric, and not DAU.
-    
+
     For more information, refer to [the DAU description in the Mozilla Data Documentation](https://docs.telemetry.mozilla.org/concepts/terminology.html#dau).
     For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
 """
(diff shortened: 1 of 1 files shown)


definitions/fenix.toml
```diff
@@ -337,46 +337,57 @@ select_expression = """
...
 select_expression = "COALESCE(SUM(activated))"
-data_source = "new_profile_activation" 
+data_source = "new_profile_activation"
 friendly_name = "New Profile Activation"
 description = "A new profile is counted as activated one week after creation if it meets the following conditions: 1) at least 3 days of use during first week 2) at least one search between days 4-7."
 owner = "vsabino@mozilla.com"
 
+[metrics.fxa_sign_in_rate]
+select_expression = """
+  COALESCE(MAX(
```
bochocki, line 382: If this is a rate, should we be using `AVG` instead of `MAX`? Could you alternatively use:
```
AVG(IF(metrics.boolean.preferences_signed_in_sync, 1, 0))
```

bochocki: Ah, that makes perfect sense. I guess one complication with Metric Hub is that it's difficult to know from the definition alone how the metric is expected to be aggregated and/or who the intended consumers are.

If we expected this to be used in a SQL query and grouped by OS/country/release_channel, I think my suggestion would make sense. But if the metric is expected to be used in experiments and aggregated by `client_id` then using `MAX` here is makes sense. I think these metrics aren't really "rates" in the way they're defined, but they are "rates" in the way they're used, which a bit confusing. Probably not a problem to solve in this PR though.

I _think_ that `MAX(IF(metrics.boolean.preferences_signed_in_sync, 1, 0))` is equivalent to the existing pattern (and is more readable imo), but maintaining consistency is also valuable; I don't have a strong preference either way.

bochocki (approved):

PR #150, opened 2023-07-06
Add sponsored tiles metrics

m-d-bowerman:
Adds defaults for sponsored tiles metrics on Desktop. This will be used for the [Third Sponsored Tiles](https://mozilla-hub.atlassian.net/browse/AD-52) experiment and for ongoing experiments for Project Honeycomb.

Adds:
Metrics:
- Sponsored tiles disabled (boolean)
- Any tile dismissals (boolean)
- Total tiles dismissals (count)
- Dismissals in positions 1 and 2 (count)
Data source:
- Sponsored tiles clients daily data source
Outcomes file:
- sponsored_tiles.toml

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -1444,7 +1444,6 @@ description = "Activity Stream Event pings"
 
 [data_sources.sponsored_tiles_clients_daily]
 from_expression = "mozdata.telemetry.sponsored_tiles_clients_daily"
-experiments_column_type = "native"
 submission_date_column = "submission_date"
 description = "Sponsored Tiles Clients Daily"
 friendly_name = "Sponsored Tiles Clients Daily"
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -143,7 +143,7 @@ description = """
 
 [metrics.unenroll]
 data_source = "normandy_events"
-select_expression='''{{agg_any(
+select_expression = '''{{agg_any(
     """
         event_category = 'normandy'
         AND event_method = 'unenroll'
@@ -216,7 +216,6 @@ friendly_name = "about:protections viewers"
 description = """
     Counts the number of clients that viewed about:protections.
 """
-
 
 
 [metrics.connect_fxa]
@@ -405,7 +404,6 @@ description = """
     Counts the number of clicks to disable Pocket sponsored content
     in New Tab made by each client.
 """
-
 
 
 [metrics.content_shutdown_crashes]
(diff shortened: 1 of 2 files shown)


definitions/firefox_desktop.toml
```diff
@@ -1255,6 +1253,43 @@ data_source = "main"
 friendly_name = "SSL Loads Probe Ratio"
 description = "Ratio of clients that have the http_pageload_is_ssl_ratio_v1 probe"
 
+[metrics.sponsored_tiles_disabled]
+select_expression = '{{agg_any("sponsored_tiles_disable_count > 0")}}'
+data_source = "sponsored_tiles_clients_daily"
+friendly_name = "Sponsored Tiles Disabled Event"
+description = "Boolean that identifies clients that disabled sponsored tiles during experiment"
```
rebecca-burwei, line 1260: Sponsored Tiles Daily:
`+` includes Mobile and Desktop
`-` based off of activity stream, which will be deprecated with pingcentre
`?` you want to use these [business filters](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/telemetry_derived/sponsored_tiles_clients_daily_v1/query.sql#L142-L165) to only return data for these clients

Newtab Interactions:
`-` Desktop only, no Mobile
`+` Consistency with other Newtab metrics: [ref 1](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L1123), [ref 2](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L1004)
`?` No business filters

```suggestion
[metrics.newtab_tiles_enabled]
data_source = "newtab_interactions"
select_expression = "COALESCE(MAX(CAST(topsites_sponsored_enabled AS INT)), 0)"
friendly_name = "Sponsored Tiles Enabled"
type = "scalar"
description = "Whether or not Sponsored tiles are enabled on the New Tab."
```

rebecca-burwei (approved):
Looks good to me. I left some suggestions in case you want to use Newtab Interactions instead of Sponsored Tiles Clients Daily, but it's up to you.

definitions/firefox_desktop.toml
m-d-bowerman: I think the business filters are fine for now, given that it covers all of the markets we're interested in. I used the `sponsored_tiles_clients_daily` source because it records the dismissal event itself, and we can measure if there's a difference in that event between branches. Although, newtab, too, would record the same, assuming the proportion of clients in each branch with that pref set is equal. Do you have thoughts on which should be used, the event or the pref?

rebecca-burwei: Actually, I'm not seeing the `topsites_sponsored_enabled` field in the newtab interactions table, even though the data is collected in the [ping](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/topsites_sponsored_enabled). I guess the probe was implemented, but not yet added to the newtab interactions table. For your Thursday deadline, let's go with Sponsored Tiles Clients Daily then.

My thought was: If the % of clients that don't have the pref set is low, then go with the pref.

PR #151, opened 2023-07-06
Release android onboarding redesign

irrationalagent:
[adds a config for this experiment](https://experimenter.services.mozilla.com/nimbus/release-android-onboarding-redesign/summary)

@danielkberry can you review this... im a bit rusty around these parts


danielkberry (approved):
I think `mozilla_default_browser` and `default_browser_action` are calculated automatically from this experiment picking the `default_browser` Outcome, so I don't think adding them here should have any effect (shouldn't hurt anything either). 

Otherwise, LGTM

PR #185, opened 2023-08-11
Add files via upload

soGaussian:
Added 4 new onboarding metrics to this file.

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -377,6 +377,35 @@ friendly_name = "New Profile Activation"
 description = "A new profile is counted as activated one week after creation if it meets the following conditions: 1) at least 3 days of use during first week 2) at least one search between days 4-7."
 owner = "vsabino@mozilla.com"
 
+
+[metrics.turn_on_notifications_onboarding]
+select_expression = "COALESCE(SUM(converted_flag))"
+data_source = "turn_on_notifications"
+friendly_name = "Turn on Notification Rate"
+description = "This metric looks at proportion of all new profiles that were exposed to the turn on notification card and completed the action during on-boarding."
+owner = "rbaffourawuah@mozilla.com"
+
+[metrics.set_to_default_onboarding]
+select_expression = "COALESCE(SUM(converted_flag))"
+data_source = "set_to_default"
+friendly_name = "Set to Default Rate"
+description = "This metric looks at proportion of all new profiles that were exposed to the set to default card and completed the action during on-boarding."
+owner = "rbaffourawuah@mozilla.com"
+
+[metrics.sign_in_onboarding]
+select_expression = "COALESCE(SUM(converted_flag))"
+data_source = "sign_in"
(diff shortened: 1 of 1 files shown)


bani:
I'm not sure this works as I'm not super familiar with the inner workings of experiment reporting.
I'm concerned this will end up in a left join to all clients in the experiment and completely ignore if they saw the card or not.
So I'll defer to @danielkberry

definitions/fenix.toml
```diff
@@ -471,3 +500,161 @@ submission_date_column = "submission_date"
...
+      SELECT
+            client_info.client_id as client_id
+            , min(DATE(submission_timestamp)) as submission_date
+      FROM
+        mozdata.org_mozilla_firefox.events tm
+      CROSS JOIN
+        UNNEST(events) AS event
+      CROSS JOIN
+        UNNEST(ping_info.experiments) AS expr
+      WHERE event.category = "onboarding" AND expr.key = '{experiment_slug}'
```
bani, line 520: metric-hub metrics also have use outside of experiments, so I wouldn't include this here

soGaussian: Sure I will remove that from the join

danielkberry:
Ok so the config is failing to compile because the SQL for querying over the events table is missing a filter over `submission_date`. Looking at the SQL, we do filter on `submission_date`, but that only gets pushed down to the `expo` subquery, not the `conv` subquery. So that will need to be changed or...

More generally, writing a bunch of data sources like this is pretty inefficient with how Jetstream puts them together (it'll have to run each query separately). Is there any way you can construct 1 data source and then write multiple metrics to query against that data source? So for example, the table should provide, at the client-day level, if they had any of the events we care about. Something like: 

```toml
[data_sources.special_onboarding_events]
from_expression = """(
SELECT 
  client_info.client_id as client_id,
  MIN(DATE(submission_timestamp)) AS submission_date,
  ANY(event.name = "set_to_default") AS set_to_default, 
  ...
  FROM 
  mozdata.org_mozilla_firefox.events 
CROSS JOIN UNNEST(events) AS event
WHERE event.name IN (...)
GROUP BY 1
)"""
```

Then you can write metrics like:

```toml
[metrics.at_least_1_cta_onboarding]
select_expression = "COALESCE(SUM(set_to_default_card + turn_on_notifications_card + sign_in_card) > 0, FALSE)"
data_source = "at_least_1_cta"
friendly_name = "Completed at least one CTA"
description = "This metric looks at proportion of all new profiles that were exposed to onboarding cards and completed at least one action during on-boarding."
owner = "rbaffourawuah@mozilla.com"
```

jetstream/on-boarding-challenge-the-default.toml
soGaussian, line 1: This needs segmentation, we wanted to segment on the attribution flag. I think we can merge and and then I will do the segmentation after the merge

danielkberry (approved):

scholtzan (approved):

PR #195, opened 2023-08-30
Adding onboarding ctas metrics to metric hub

soGaussian:
Added all the onboarding CTAs to metric hub across desktop, android and firefox. 
All the data sources follow the same logic. Desktop data source was completely built by their own engineers using the fennix example.
I had to make slight change to the fenix file to remove any ambiguity with regards to the event.


soGaussian:
Going to submit PR separately for each component so I move fater

PR #206, opened 2023-09-05
Ios ctd onboarding

soGaussian:
I updated the logic in the definition file and created overrides for the enrollment dates


mikewilli (changes requested):
My only concern here is that these metrics may be used in other experiments, so changing their identifiers might cause issues elsewhere. Can you confirm that this experiment (the one for which you've added a custom config on the PR) is the only place this metric is used? If not, it would be safest to leave the name unchanged and only change the `friendly_name` (or you could duplicate the metric and rename the new one).

mikewilli (approved):

PR #212, opened 2023-09-12
Use legacy client_id in newtab_interactions

m-d-bowerman:
Currently, the `newtab_interactions` data source uses the Glean `client_id` column. This will result in all 0 values for `newtab_interactions` metrics when Jetstream attempts to join to the enrollments table.


rebecca-burwei (approved):
lgtm

PR #213, opened 2023-09-12
[ci skip-rerun] Remove split search engine as a default metric for desktop

danielkberry:


danielkberry:
This PR will remove the `separate search engine` as a default metric for desktop. 
@scholtzan : my intention is to apply this on a going-forward basis (i.e., new experiments launched after this will not get this metric, currently live experiments will get it). Does this PR accomplish that?
@mikewilli will this have any impacts on experimenter or will it adapt properly?

mikewilli:
You should be good on both. 

The default behavior for metric changes (i.e., "do nothing" in the [docs](https://experimenter.info/jetstream/jetstream/#how-to-use-the-latest-tooling-and-metric-definitions)) is the behavior you described -- new experiments will get the new metrics, existing experiments are unaffected.

And since this isn't a guardrail, the metric should just not appear on the results page because it won't exist in the jetstream results.

definitions/firefox_desktop.toml
```diff
@@ -1290,6 +1290,11 @@ data_source = "activity_stream_events"
 friendly_name = "Sponsored Tiles Dismissals Count (Positions 1 and 2)"
 description = "Count of sponsored tiles dismissals in the first two positions"
 
+[metrics.separate_search_engine]
```
mikewilli, line 1293: Where was this defined before?

mikewilli (approved):

definitions/firefox_desktop.toml
danielkberry: I forgot to commit the change removing the definition. Previously, this was defined in [jetstream/defaults/firefox_desktop.toml](https://github.com/mozilla/metric-hub/pull/213/files#diff-9db7038fbcff2586e8a7ca3185f7e6e67eb301454499a7cf59dcf011dcc11e03L59-L65), which is the wrong place for a definition to live.

scholtzan (approved):

PR #218, opened 2023-09-13
Added http download and upload throughput metrics to Networking Outcome

acreskeyMoz:
Adding histograms,
```
networking_download_throughput_http_1
networking_download_throughput_http_2
networking_download_throughput_http_3
```

And keyed histogram `http_upload_bandwidth_mbps` with keys,
```
http/1.1
h2
h3
```

### DIFF
--- a/jetstream/outcomes/firefox_desktop/networking.toml
+++ b/jetstream/outcomes/firefox_desktop/networking.toml
@@ -2,7 +2,7 @@ friendly_name = "Networking"
 description = "Network layer metrics including Necko, TLS, NSS"
 
 [metrics.dns_lookup_time]
-select_expression = '{{agg_histogram_mean("payload.processes.content.histograms.dns_lookup_time")}}'
+select_expression = '{{agg_histogram_mean("payload.histograms.dns_lookup_time")}}'
 data_source = 'main'
 bigger_is_better = false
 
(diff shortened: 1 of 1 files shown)


jetstream/outcomes/firefox_desktop/networking.toml
```diff
@@ -2,7 +2,7 @@ friendly_name = "Networking"
 description = "Network layer metrics including Necko, TLS, NSS"
 
 [metrics.dns_lookup_time]
-select_expression = '{{agg_histogram_mean("payload.processes.content.histograms.dns_lookup_time")}}'
+select_expression = '{{agg_histogram_mean("payload.histograms.dns_lookup_time")}}'
```
acreskeyMoz, line 5: ^ This is an unrelated bug fix.

acreskeyMoz:
That's good point.

I'm in the process of making a new outcome for metrics that are specific to a given experiment (Early Hints).

And let me do the same for these metrics, upload and download -- they don't need to be calculated for _every_ networking outcome.

acreskeyMoz:
@dennisjackson what do you think about moving some of the TLS specific metrics into a separate outcome?

acreskeyMoz:
upload / download moved into new outcome: https://github.com/mozilla/metric-hub/pull/219

PR #225, opened 2023-09-19
Add the Early Hints-specific outcome

acreskeyMoz:
This consists of the keyed histograms: `eh_perf_page_load_time_ms` and `eh_perf_first_contentful_paint_ms`. For each, we are interested in the `preconnect_` key.


danielkberry (approved):

danielkberry:
As per conversations with @scholtzan & @mikewilli , the last failing check is expected for PRs made from forks and can be ignored

PR #237, opened 2023-10-09
Garbage collection

lelilia:
This is my first try with metric hub, so please give me all the feedback :) 

There is no real use case yet and the idea is mostly to get a feel for OpMon and go from there. 

I did not find the metric `gc_effectiveness` in the defaults and I'm not sure if it would be better to add it to the firefox_desktop defaults or put it here as a one off ...

### DIFF
--- a/opmon/garbage_collection.toml
+++ b/opmon/garbage_collection.toml
@@ -1,0 +1,54 @@
+[project]
+
+name = "Garbage Collection"
+platform = "firefox_desktop"
+xaxis = "submission_date"
+start_date = "2023-10-07"
+skip = false
+is_rollout = false
+skip_default_metrics = true
+compact_visualization = false
+
+metrics = [
+    "gc_ms",
+    "gc_budget_overrun",
+    "gc_effectiveness"
+]
+
+alerts = [
+    "historical_diff"
+]
+
+[project.population]
+
+data_source = "main"
+# channel = "release"
+channel = "beta"
+monitor_entire_population = true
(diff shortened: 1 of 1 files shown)


opmon/garbage_collection.toml
```diff
@@ -0,0 +1,53 @@
...
+]
+
+alerts = [
+    "historical_diff"
+]
+
+[project.population]
+
+data_source = "main"
+channel = "release"
```
scholtzan, line 25: Just FYI, by default this will run on 1% of the release data to save cost

scholtzan (approved):

opmon/garbage_collection.toml
lelilia: So I should remove the next line `monitor_entire_population = true` which would overwrite this default? or is that unconnected?

scholtzan: No, that's unrelated. The `main` data source is querying from `main_1pct`: https://github.com/mozilla/metric-hub/blob/3f7fdf3cc993ffd5f6557dd51313a6d43297663b/opmon/defaults/firefox_desktop.toml#L50

So if you really (really) need it then you can override the `main` data source and query from `main_v5` instead. 
Do you think it would make a difference for these metrics?

lelilia: no I really don't need more than 1pct ... just wanted to make sure that I don't accidentally query everything 
thank you very much

PR #238, opened 2023-10-09
[ci skip-rerun] Adds DAU as a metric

danielkberry:
Adds initial attempt for fx desktop

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -457,6 +457,7 @@ from_expression = """(
 )"""
 client_id_column = "client_info.client_id"
 experiments_column_type = "glean"
+submission_date_column = "submission_date"
 default_dataset = "fenix"
 friendly_name = "Baseline"
 description = "Baseline Ping"
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -456,7 +456,6 @@ from_expression = """(
     FROM `moz-fx-data-shared-prod.{dataset}.baseline` p
 )"""
 client_id_column = "client_info.client_id"
-submission_date_column = "DATE(submission_timestamp)"
 experiments_column_type = "glean"
 default_dataset = "fenix"
 friendly_name = "Baseline"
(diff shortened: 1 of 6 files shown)


scholtzan (approved):

scholtzan:
The reason for CI failing is here: https://github.com/mozilla/metric-hub/blob/9a6bcc770033c31bbc952a70ba82551d577c959c/definitions/firefox_ios.toml#L370

This should just be `submission_date`

danielkberry:
@bochocki , this PR will add DAU as an experiment guardrail, can you review? This does not replace DAU Impact, but should replace QCDOU in the experiment results.

bochocki (approved):
This looks good to me! My only note is that there's a very good chance we'll update the mobile DAU definition in the not-so-distant future, so the fenix and firefox_ios files will need to be updated accordingly.

danielkberry:
@bochocki when that update happens, would you be landing a new version of `client_level_daily_active_users_v1` as `client_level_daily_active_users_v2`?

bochocki:
@danielkberry I will if I'm around, but it's unclear to me what the timeline for the definition change is (although I think it's likely to happen in Q4).

I think it makes sense to write a new definition called `client_level_daily_active_users_v2`, which will probably look something like:
```
COUNT(DISTINCT IF(
    LOWER(COALESCE(metadata.isp.name, "")) <> 'browserstack'
    AND COALESCE(metrics.timespan.glean_baseline_duration.value, 0) > 0
    , client_info.client_id, NULL
))
```

danielkberry:
@bochocki awesome, thanks. I think here we're somewhat agnostic to the actual definition, more want to confirm that the definitions are versioned and we land a new version instead of modifying the `_v1` version. That way we can ensure experiment analyses are consistent (an experiment gets either the old definition or the new definition, but not a mixture of both).

danielkberry:
@scholtzan before I merge this, can you confirm that it will not retrigger/reset dates for live experiments and only new experiments will pick it up?

scholtzan:
yup, can confirm `[ci skip-rerun]` should handle that

PR #239, opened 2023-10-09
Create a minimal networking performance outcome

acreskeyMoz:
Just the performance metrics from the Networking outcome.

### DIFF
--- a/jetstream/outcomes/firefox_desktop/networking_performance.toml
+++ b/jetstream/outcomes/firefox_desktop/networking_performance.toml
@@ -1,5 +1,5 @@
 friendly_name = "Networking Performance"
-description = "Network layer performance metrics"
+description = "Network layer performance metrics (subset of Networking outcome without tls and error metrics)"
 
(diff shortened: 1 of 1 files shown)


jetstream/outcomes/firefox_desktop/networking_performance.toml
```diff
@@ -0,0 +1,86 @@
+friendly_name = "Networking Performance"
+description = "Network layer performance metrics"
```
mikewilli, line 2: I'd suggest something that indicates that this is a subset of the other outcome.
```suggestion
description = "Network layer performance metrics (subset of Networking outcome without JS performance metrics)"
```

mikewilli (approved):
One comment but LGTM, thanks!

jetstream/outcomes/firefox_desktop/networking_performance.toml
acreskeyMoz: Perfect, I've updated it with a comment regarding the subseting.

PR #240, opened 2023-10-10
funnel metrics definition in metric hub

soGaussian:
This is to merge funnel metrics definition in metrics hub

### DIFF
--- a/definitions/firefox_ios.toml
+++ b/definitions/firefox_ios.toml
@@ -289,7 +289,7 @@ owner = "rbaffourawuah@mozilla.com"
 deprecated = false
 
 
-[metrics.new_profiles]
+[metrics.funnel_new_profiles]
 data_source = "funnel_retention"
 select_expression = "SUM(new_profiles)"
 type = "scalar"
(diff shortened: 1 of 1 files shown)


definitions/firefox_ios.toml
```diff
@@ -260,6 +260,77 @@ friendly_name = "Clicked at least one CTA"
...
+    This is the total number of downloads of firefox browser in iOS appstore. The etl of the base table is
+    [defined in `bigquery-etl`](https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_ios_derived/app_store_funnel_v1/query.sql).
+    This metric needs to be aggregated by `first_seen_date` (date column from the data recieved from appstore) for daily aggregation. The underlying table have a lag of 7 days.
+    For questions please contact "rbaffourawuah@mozilla.com".
+"""
+owner = "rbaffourawuah@mozilla.com"
+deprecated = false
+
+
+[metrics.new_profiles]
```
bani, line 292: My suggestions is to call this funnel_new_profiles, to avoid confusion with what we normally call new_profiles, which even though isn't defined in metric hub yet, it should have been, pointing to the firefox_ios_clients table.

This is particularly important once we create the same metrics on Android, which we want to keep consistent, as the Android funnel new profiles is a very small subset of total new profiles.

The same comment is applicable to the other retention metrics

bani (changes requested):

definitions/firefox_ios.toml
soGaussian: Updated the new_profiles name

bani (approved):

PR #244, opened 2023-10-11
[ci rerun-skip] add monitor_cirrus datasource and metric

mikewilli:
I'm not sure what to try measuring, but I wanted to at least get something out there for feedback. 

- Does this look broadly correct?
- @danielkberry can you help with what the metric definition should actually be?


danielkberry (approved):
I think the page views metric is good enough for now. Ilana & Kimmy will be able to give real metrics soon

PR #245, opened 2023-10-12
subset of outcome metrics for `speculative-connect-sockets-increased-population`

mikewilli:
Workaround because custom experiment configs do not currently support overriding outcomes. Instead, we list out all of the metrics from the outcomes on the experiment, and set the `select_expression` to a no-op for metrics we don't want to compute.

### DIFF
--- a/jetstream/speculative-connect-sockets-increased-population.toml
+++ b/jetstream/speculative-connect-sockets-increased-population.toml
@@ -1,0 +1,125 @@
+[experiment]
+
+## Metrics
+[metrics]
+
+overall = [
+    'dns_lookup_time',
+    'http_page_tls_handshake',
+    'http_sub_tls_handshake',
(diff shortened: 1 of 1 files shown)


scholtzan (approved):
Probably worth testing on a single day before we rerun everything just to realize that this doesn't work 😬

mikewilli:
One thought I had while creating this -- wouldn't just not including the metrics in any of `[overall, weekly, daily]` exclude them from analysis? Or does the outcome definition include that and overrides what we have here?

mikewilli:
This is blocked on https://github.com/mozilla/metric-config-parser/pull/335, so waiting for that and a subsequent Jetstream PR before merging.

mikewilli:
This is good to merge now with https://github.com/mozilla/jetstream/pull/1988

PR #248, opened 2023-10-17
DENG-1612 - Added new_profiles metric

ANich:
Adds minimal `new_profiles` metric


mikewilli (approved):
LGTM

irrationalagent (approved):
Looks good for now, with the caveat that if we find out a reliable way to remove any possible spurious clients we might want to update this, or add an additional more strict definition.

PR #255, opened 2023-11-02
add enrollment end date and period for monitor-test-production

mikewilli:
Cirrus experiments require some updates to Experimenter, so in the meantime for testing purposes, this adds a retroactive end to enrollment for the monitor experiment.


danielkberry (approved):
LGTM

PR #262, opened 2023-11-10
Add unnested version of Urlbar Events table

dzeber:
Unnesting the results makes it more convenient to define metrics filtering on a result type.


rebecca-burwei (approved):
lgtm

PR #263, opened 2023-11-13
fix for scalar subquery produced more than one element

mikewilli:
This experiment has errors in analysis like `Scalar subquery produced more than one element`, so I rearranged the scalar subquery in the metrics causing the errors. Please take a look in case something looks off.

Also just wanted to confirm: are `'bookmarks-bar_pinned_on'` and `'bookmarks-bar_pinned_off'` typos? It seems odd to have a mix of hyphens and underscores, but I don't know what these keys normally look like so wanted to ask.

### DIFF
--- a/jetstream/bookmarks-toolbar-default-on.toml
+++ b/jetstream/bookmarks-toolbar-default-on.toml
@@ -5,12 +5,12 @@ overall = ['bookmarks_bar_pinned_on', 'bookmarks_bar_pinned_off', 'bookmarks_bar
 
 
 [metrics.bookmarks_bar_pinned_on]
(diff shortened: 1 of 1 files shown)


florezita:
Hi Mike! This my first PR review requested to it didn't register I was to review it. 'bookmarks-bar_pinned_on' and 'bookmarks-bar_pinned_off' aren't typos, that's how those probes are keyed.

Thank you for tackling this-- the PR passed checks and merged so I figured things were okay.

mikewilli:
Great thanks! For both the review and confirming the probe keys are correct. And yea the PR checks are helpful but not perfect, hopefully this time around all will go smoothly.

PR #268, opened 2023-11-15
DENG-1881 OKR 2.6 Performance Review Dashboard Kenya

lelilia:
https://mozilla-hub.atlassian.net/browse/DENG-1881

not finished yet, need to find out how to do the http custom distributions, but would appreciate a look on the rest to see if I'm on the right track

I want to compare Kenya to the US so I created a new dataset filtering for those to countries


scholtzan:
mmmh, there might be something wrong with the validation? Somehow it is looking for an experiment related to this config...

lelilia:
yeah that as well ... no idea why it's doing this?

But another question: I want to compare 2 countries ... I've seen looker dashboards that compare 2 things [link](https://mozilla.cloud.looker.com/dashboards/operational_monitoring::bug_1759000_rollout_initial_rollout_of_tcp_release_91_99?Date=&Percentile=50) I can't find the code that build it but I assume this is due to it being a rollout ... is there a way to say compare my dimensions? So it's not a filter but those two things? that would be super useful

Could I just merge this and see what I've got so far, or should I try to find out why it's looking for a experiment?

scholtzan:
One way to compare two countries is to defined `metric_groups` (similar to this https://github.com/mozilla/metric-hub/blob/b826770682438a1c2bd4fa8a285e2f33bca41f70/opmon/firefox-install-demo.toml#L29)

You'd need to define a metric for the both countries separately though

opmon/performance-review-kenya.toml
```diff
@@ -0,0 +1,469 @@
...
+[data_sources]
+
+[data_sources.baseline_US_and_KE]
+
+from_expression = """(
+    SELECT
+        p.*,
+        DATE(p.submission_timestamp) AS submission_date
+    FROM `moz-fx-data-shared-prod.{dataset}.baseline` p
+    WHERE metadata.geo.country in ("US", "KE")
```
scholtzan, line 161:
```suggestion
    WHERE metadata.geo.country in ("US", "KE")
    AND sample_id >=0 AND sample_id < 10
```
maybe a 10% sample could be enough here

opmon/performance-review-kenya.toml
```diff
@@ -0,0 +1,469 @@
...
+friendly_name = "Baseline"
+description = "Baseline Ping"
+build_id_column = "REPLACE(CAST(DATE(mozfun.norm.fenix_build_to_datetime(client_info.app_build)) AS STRING), '-', '')"
+
+
+[metrics]
+
+# Storage Stats App Bytes
+[metrics.storage_stats_app_bytes_KE]
+data_source = "metrics"
```
scholtzan, line 176: it might also help to sample the `metrics` data source since this will be joined onto the population

PR #269, opened 2023-11-16
Replace activity stream metrics with newtab

m-d-bowerman:
Replaces the metrics in `pocket_newtab.toml` with New Tab equivalents, where available. Currently, there is not a PREF_CHANGED event instrumented in `newtab`, so I've replaced those in the outcome file with metrics that record the pref. When the event is instrumented, these should be swapped over.

### DIFF
--- a/jetstream/outcomes/firefox_desktop/pocket_newtab.toml
+++ b/jetstream/outcomes/firefox_desktop/pocket_newtab.toml
@@ -1,72 +1,14 @@
 friendly_name = "Pocket in New Tab"
 description = "Usage and engagement metrics for Pocket content in New Tab."
 
-[metrics.pocket_organic_rec_clicks]
-friendly_name = "Pocket organic rec clicks"
-description = "Number of Pocket organic recs clicked in New Tab"
-select_expression = """
-      COUNTIF(
-          event = 'CLICK'
-          AND source = 'CARDGRID'
-          AND JSON_EXTRACT_SCALAR(value, '$.card_type') = 'organic'
-      )
-"""
-data_source = "as_events"
-statistics = { bootstrap_mean = {} }
+[metrics.organic_pocket_clicks.statistics.bootstrap_mean]
+[metrics.organic_pocket_clicks.statistics.deciles]
 
-[metrics.pocket_sponsored_content_clicks]
-friendly_name = "Pocket sponsored content clicks"
-description = "Number of Pocket sponsored content tiles clicked in New Tab"
-select_expression = """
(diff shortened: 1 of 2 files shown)


sergiosonline (commented):
@m-d-bowerman : Looks good from my end. @kirill-demtchouk : Are you comfortable swapping to newtab for these Pocket indicators? I know there were some outstanding questions in the Activity Stream validation work and wanted to make sure this swap doesn't impact our ability to perform inference.

> Currently, there is not a PREF_CHANGED event instrumented in newtab, so I've replaced those in the outcome file with metrics that record the pref. When the event is instrumented, these should be swapped over.

@m-d-bowerman : Would adding this instrument to newtab unlock any processing or accuracy benefit, either in the context of analyzing experiments or reporting? We can add this ask into our newtab telemetry backlog.

jetstream/outcomes/firefox_desktop/pocket_newtab.toml
```diff
@@ -1,72 +1,14 @@
...
-      COUNTIF(
-          event = 'CLICK'
-          AND source = 'CARDGRID'
-          AND JSON_EXTRACT_SCALAR(value, '$.card_type') = 'organic'
-      )
-"""
-data_source = "as_events"
-statistics = { bootstrap_mean = {} }
+[metrics.organic_pocket_clicks.statistics.bootstrap_mean]
+[metrics.organic_pocket_clicks.statistics.deciles]
```
kirill-demtchouk, line 5: This is cool! My assumptions is that this will test for differences in the distribution of Pocket clicks per client ID? How does partybal visualize such a thing?

kirill-demtchouk (approved):

jetstream/outcomes/firefox_desktop/pocket_newtab.toml
m-d-bowerman: Yup, for differences in mean clicks per client. We'll get results on Experimenter that show confidence intervals for the means of the branches, as well as an interval for relative uplift in the metric due to the treatment. Partybal shows similar results, but it's where you'd have to look if segments were included -- Experimenter doesn't show results for segments yet.

m-d-bowerman:
@sergiosonline -- yeah, currently we only know the state of that pref, but it'd be useful for reporting if we knew exactly when the pref was flipped. It's not a huge deal, as we know that disables are a very rare event. Chutten implemented the event probe for tiles, and instrumenting the same thing for Pocket would be an almost identical change.

PR #270, opened 2023-11-17
add end date in line with original duration

mikewilli:
This experiment was still running until today despite ending being requested. Adding a custom config so that Jetstream doesn't needlessly compute 6 months worth of data.


danielkberry (approved):
LGTM

PR #295, opened 2023-12-11
Review checker reset end date

danielkberry:


danielkberry:
As per this [PR](https://github.com/mozilla/metric-hub/pull/284), I'm reverting the date change now that the results readout has passed and we need final numbers instead of preliminary numbers.

PR #304, opened 2023-12-21
Add metrics to PDF outcome

danielkberry:


danielkberry:
As per [the docs](https://experimenter.info/deep-dives/jetstream/outcomes/#what-happens-if-an-outcome-changes) this PR should be a no-op for live experiments, so should be safe to merge at this time.

PR #332, opened 2024-01-18
change search data source filters and fix fenix uri count

irrationalagent:
We noticed that the search metrics for Fenix in jetstream were broken. Using the current definition (pasted below) now returns no results:

```sql
(
    SELECT *
    FROM `mozdata.search.mobile_search_clients_engines_sources_daily`
    WHERE normalized_app_name = "Fenix"
)
```

This is because "Fenix" is no longer a value in `normalized_app_name`. I'm not sure when or where this changed, but it appears to be recent. The filter needs to be changed to either:

```sql
WHERE app_name = "Fenix"
```
OR 
```sql
WHERE normalized_app_name_os = "Firefox Android"
```

[It appears that these two filters return the same data](https://sql.telemetry.mozilla.org/queries/97049/source).

I also went ahead and changed the filters for the other apps to follow the same pattern. 

I also changed the URI count metric on fenix to point to a non-expired probe. [The current metric](https://github.com/mozilla/metric-hub/blob/main/jetstream/defaults/fenix.toml#L99-L101)  [points to an expired probe](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/events_total_uri_count). I believe it should [use this one instead](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/events_normal_and_private_uri_count).

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -2,7 +2,7 @@
 
 [metrics.uri_count]
 data_source = "baseline"
-select_expression = '{{agg_sum("metrics.counter.events_total_uri_count")}}'
+select_expression = '{{agg_sum("metrics.counter.events_normal_and_private_uri_count")}}'
 friendly_name = "URIs visited"
 description = "Counts the number of URIs each client visited"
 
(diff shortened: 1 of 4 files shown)


xluo-ds:
Yeah, there is recently a change to the `normalized_app_name` in the table you cited above. I'm adding @pissac17  to review

pissac17 (approved):
LGTM, here's the [mapping table](https://docs.google.com/document/d/15ZPQKPudbeX30klYQDpLjJbsVfdebgjhxmUgRVLNPyY/edit) for future reference

PR #343, opened 2024-01-29
[ci skip-rerun] Update DOU to use new DAU qualifiers for mobile

danielkberry:


mikewilli:
@danielkberry Can this be closed?

danielkberry:
I think so, the updated definition is in the client level daily active users metric, which is what we intend to show on the UI

PR #358, opened 2024-02-12
Update macos-background-tab-power-savings.toml

danielkberry:


jetstream/macos-background-tab-power-savings.toml
```diff
@@ -93,5 +90,41 @@ type = "histogram"
...
+
+[data_sources]
+[data_sources.main_filtered]
+from_expression = """(
+    SELECT
+        *,
+        DATE(submission_timestamp) AS submission_date,
+        environment.experiments
+    FROM `moz-fx-data-shared-prod.telemetry_stable.main_v5`
+    WHERE mozfun.map.get_key(environment.experiments, 'macos-background-tab-power-savings') IS NOT NULL
```
scholtzan, line 126: filtering on the beta channel might speed up things even more here (afaik this experiment is running on beta)

danielkberry: Can we do that automatically when building these queries? I believe channel should be in the Experimenter API

scholtzan: it's a little more complicated. I opened https://github.com/mozilla/jetstream/issues/1827 a while ago which mentions this as well

PR #366, opened 2024-02-22
fix onboarding metric definition

mikewilli:
These metrics are defined in the Default Browser outcome as well, and one of the definitions in Onboarding are out of date. To avoid this in the future, we should remove the duplicate definitions so they can be updated in only one place.


mikewilli:
I tried removing the definitions because they are duplicated in Default Browser, but this produced a CI error because it couldn't find the definition. For now, let's just fix the definition and worry about a better fix later.

PR #367, opened 2024-02-29
Demo fenix feature usage

scholtzan:
Some metric for demoing automatic Looker artifact generation

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -619,12 +619,12 @@ experiments_column_type = "none"
 from_expression = """(
     SELECT
         *
-    FROM `moz-fx-data-shared-prod.fenix.feature_usage_metrics` p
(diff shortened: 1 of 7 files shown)


definitions/fenix.toml
```diff
@@ -585,3 +613,23 @@ GROUP BY 1, 2, 3, 4, 5, 6
 )
 """
 experiments_column_type = "none"
+
+
+[data_sources.feature_usage_metrics_v1]
+from_expression = """(
+    SELECT
+        *
+    FROM `moz-fx-data-shared-prod.fenix.feature_usage_metrics` p
```
kik-kik, line 622: Should we be using mozdata instead of the `moz-fx-data-shared-prod` GCP project here for views exposed to mozdata??

scholtzan: yup, good call

PR #368, opened 2024-03-01
feat(DENG-2431): Update new_profile_activation source definition to point to the new activation table

kik-kik:
# feat(DENG-2431): Update new_profile_activation source definition to point to the new activation table

`firefox_ios.new_profile_activation` has been replaced by a more reliable `firefox_ios.clients_activation`. This is just a follow up to that change to make sure the correct table is used.

blocks: https://github.com/mozilla/bigquery-etl/pull/5149

### DIFF
--- a/definitions/firefox_ios.toml
+++ b/definitions/firefox_ios.toml
@@ -250,7 +250,7 @@ description = "Number of times Contile Sponsored Tiles setting is flipped."
 owner = "xluo-all@mozilla.com"
 
 [metrics.new_profile_activation]
-select_expression = "COALESCE(SUM(activated))"
+select_expression = "COUNTIF(is_activated)"
 data_source = "new_profile_activation"
 friendly_name = "New Profile Activation"
 description = "A new profile is counted as activated one week after creation if it meets the following conditions: 1) at least 3 days of use during first week 2) at least one search between days 4-7."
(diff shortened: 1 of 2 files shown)


definitions/firefox_ios.toml
```diff
@@ -441,7 +441,7 @@ client_id_column = "client_id"
 submission_date_column = "submission_date"
 
 [data_sources.new_profile_activation]
-from_expression = "`moz-fx-data-shared-prod.firefox_ios.new_profile_activation`"
```
kik-kik, line 444: Alternatively, we could use `firefox_ios.firefox_ios_clients` here instead as the views also contains this field:
https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_ios/firefox_ios_clients/view.sql#L21

bani (approved):

definitions/firefox_ios.toml
```diff
@@ -250,7 +250,7 @@ description = "Number of times Contile Sponsored Tiles setting is flipped."
 owner = "xluo-all@mozilla.com"
 
 [metrics.new_profile_activation]
-select_expression = "COALESCE(SUM(activated))"
+select_expression = "COUNTIF(is_activated)"
```
scholtzan, line 253: @danielkberry @mikewilli  should we create a new metric here, or change the existing one in place?

PR #370, opened 2024-03-06
Firefox set-as-default (UserChoice) metrics.

ncalexan:


ncalexan:
I verified the various statistics by hand in R to ensure that parts added up, finding more than one typo in the process.

I would very much appreciate hearing about alternate approaches, because unrolling a categorical histogram by hand like this is not pleasant.

scholtzan:
cc @lelilia  maybe you have some ideas of how this could be improved, but at least for awareness

I think for now there isn't a better or more convenient approach, unfortunately.

PR #371, opened 2024-03-07
Scale Firefox set-as-default (UserChoice) metrics by 100 to account for `sample_id`

ncalexan:
I overlooked this before landing.

While we're here, can we schedule this to backfill from `2023-01-01`, like it says in the TOML?  I expected that to happen last night.


scholtzan (approved):

scholtzan:
I'll start a backfill. Right now these need to be triggered manually

PR #372, opened 2024-03-12
Pages ranking experiment

teonbrooks:
Fixed some old links.

Added a configuration file for a new experiment.

### DIFF
--- a/jetstream/README.md
+++ b/jetstream/README.md
@@ -14,8 +14,8 @@ Additional reviews by data scientists are necessary when making changes to [metr
 
 Are your analyses not rendering as expected? See https://experimenter.info/jetstream/troubleshooting.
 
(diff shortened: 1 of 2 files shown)


pissac17 (changes requested):
@teonbrooks looks like `history clicks` and `history impression `have the same SQL logic. I see the same for `search suggestion clicks` and `search suggestion  impressions` too.

jetstream/pages-ranking-experiment.toml
```diff
@@ -0,0 +1,24 @@
+[experiment]
```
pissac17: You could  remove the [experiment] section with the dates. It'll automatically take the dates specified in the experimenter.

pissac17 (approved):
LGTM

PR #375, opened 2024-03-18
[ci skip-rerun] fix: update jetstream defaults to use non-deprecated search count

### LINKED ISSUE #34: Fenix tagged searches default metric for Jetstream uses deprecated definition, migrate to newer version
The definition [here](https://github.com/mozilla/metric-hub/blob/main/definitions/fenix.toml#L166) is marked as deprecated but used as default [here](https://github.com/mozilla/metric-hub/blob/main/jetstream/defaults/fenix.toml#L8)

jaredsnyder:
fixes #34 

[ci skip-rerun]

### DIFF
--- a/jetstream/defaults/fenix.toml
+++ b/jetstream/defaults/fenix.toml
@@ -5,7 +5,7 @@ weekly = [
     "active_hours",
     "search_count",
     "serp_ad_clicks",
-    "tagged_sap_searches",
+    "tagged_search_count",
     "total_uri_count",
     "days_of_use",
     "client_level_daily_active_users_v2",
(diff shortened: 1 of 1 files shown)


danielkberry:
@jaredsnyder what are your thoughts on adding the correct iOS metric as a default metric at the same time. I [see](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_ios.toml#L181-L194) there's already the right metric defined, should that be a default metric also?

danielkberry:
> > @jaredsnyder what are your thoughts on adding the correct iOS metric as a default metric at the same time. I [see](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_ios.toml#L181-L194) there's already the right metric defined, should that be a default metric also?
> 
> I do not feel qualified to answer but I will ask the revenue team

Ok, then let's treat that as a separate issue and not block on it for this PR

danielkberry (approved):
LGTM

PR #382, opened 2024-04-01
Point new tab metrics to clients daily

m-d-bowerman:
Update new tab metrics that are not position-based to pull from `newtab_clients_daily`


m-d-bowerman:
I'm meeting with DE tomorrow to do the backfill of `newtab_visits` and `newtab_clients_daily`. I'll submit this PR when that is complete to avoid complications when using Mozanalysis for sizing with New Tab metrics where no data yet exists.

PR #391, opened 2024-04-26
[RS-1063] Adding custom config for 'sponsored-suggestion-placement-phase-2'

sergiosonline:
Read through https://mozilla.slack.com/archives/CF94YGE03/p1711033682724839 and some of the relevant Experimenter docs. This should be in the spirit of producing two cohorts for analysis.

* Ran `jetstream validate_config ...`, but it's giving the issue and I'd like y'all's help to debug:
```Column name enrollment_date is ambiguous```

### DIFF
--- a/jetstream/sponsored-suggestion-placement-phase-2.toml
+++ b/jetstream/sponsored-suggestion-placement-phase-2.toml
@@ -8,11 +8,11 @@ segments = ["cohort_1", "cohort_2"] # Two cohorts
 
 [segments]
 [segments.cohort_1]
-select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-01-30'"
+select_expression = "segment_enrollment_date BETWEEN '2024-01-17' AND '2024-01-30'"
 data_source = "segment_enrollments"
 
 [segments.cohort_2]
-select_expression = "enrollment_date BETWEEN '2024-01-31' AND '2024-02-20'"
+select_expression = "segment_enrollment_date BETWEEN '2024-01-31' AND '2024-02-20'"
 data_source = "segment_enrollments"
 
 [segments.data_sources.segment_enrollments]
--- a/jetstream/sponsored-suggestion-placement-phase-2.toml
+++ b/jetstream/sponsored-suggestion-placement-phase-2.toml
@@ -1,0 +1,694 @@
+[experiment]
+
+enrollment_period = 14
+end_date = '2024-04-22'
+segments = ["cohort_1", "cohort_2"] # Two cohorts
+
+## Segments
+
+[segments]
+[segments.cohort_1]
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-01-30'"
+data_source = "segment_enrollments"
+
+[segments.cohort_2]
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-02-06'"
+data_source = "segment_enrollments"
+
+[segments.data_sources.segment_enrollments]
+client_id_column = "client_id"
+submission_date_column = "enrollment_date"
+friendly_name = "Enrolment table used to calculate above segments"
+from_expression = """(
+    SELECT
+        e.client_id,
+        `mozfun.map.get_key`(e.event_map_values, 'branch')
+            AS branch,
+        MIN(e.submission_date) AS enrollment_date,
+        COUNT(e.submission_date) AS num_enrollment_events
+    FROM `moz-fx-data-shared-prod.telemetry.events` e
+    WHERE
+        e.event_category = 'normandy'
+        AND e.event_method = 'enroll'
+        AND e.submission_date
+            BETWEEN '2024-01-16' AND '2024-04-22'
+        AND e.event_string_value = 'sponsored-suggestion-placement-phase-2'
+    GROUP BY e.client_id, branch
+)"""
+
+[metrics]
+
+overall = [
+    # Legacy SERP
+    "legacy_serps_with_ads",
+    "legacy_ad_clicks",
+    "adm_impressions",
+    # Glean
+    "adm_clicks",
+    "adm_impression_rate",
+    "adm_ctr",
+    "serp_impressions",
+    "serp_impressions_with_ads",
+    "serp_has_ad_click",
+    "serp_ad_clicks",
+    "serp_ad_impression_rate",
+    "serp_ad_ctr",
+    "serp_ad_click_rate",
+    # Glean session-based
+    "urlbar_sessions",
+    "urlbar_clicks_per_session",
+    "serps_per_session",
+    "serps_with_ads_per_session",
+    "ad_clicks_per_session",
+    "adm_impressions_per_session",
+    "adm_clicks_per_session",
+    # Exposed metrics
+    "exposed_urlbar_sessions",
+    "exposure_rate",
+    "exposed_search_engine_clicks",
+    "exposed_search_engine_rate",
+    "exposed_serp_impressions",
+    "exposed_serp_impressions_with_ads",
+    "exposed_serp_has_ad_click",
+    "exposed_serp_ad_clicks",
+    "exposed_serp_ad_impression_rate",
+    "exposed_serp_ad_ctr",
+    "exposed_serp_ad_click_rate",
+    # Exposed session-based
+    "exposed_urlbar_clicks_per_session",
+    "exposed_serps_per_session",
+    "exposed_serps_with_ads_per_session",
+    "exposed_ad_clicks_per_session",
+    "exposed_adm_impressions_per_session",
+    "exposed_adm_clicks_per_session",
+]
+
+
+#----------------------------------------------
+
+# Legacy SERP metrics
+
+[metrics.legacy_serps_with_ads]
+select_expression = """{{agg_sum("CASE WHEN normalized_engine = 'Google' THEN search_with_ads ELSE 0 END")}}"""
+data_source = "search_clients_engines_sources_daily"
+friendly_name = "Legacy SERPs with ads"
+description = "Tagged Google SERP page loads with ads loaded from Legacy Telemetry"
+
+[metrics.legacy_serps_with_ads.statistics.bootstrap_mean]
+[metrics.legacy_serps_with_ads.statistics.deciles]
+
+[metrics.legacy_ad_clicks]
+select_expression = """{{agg_sum("CASE WHEN normalized_engine = 'Google' THEN ad_click ELSE 0 END")}}"""
+data_source = "search_clients_engines_sources_daily"
+friendly_name = "Legacy ad clicks"
+description = "Ad clicks on tagged Google SERPs from Legacy Telemetry"
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: jetstream/sponsored-suggestion-placement-phase-2.toml [hunk 1]
```diff
@@ -0,0 +1,694 @@
+[experiment]
+
+enrollment_period = 14
```
danielkberry, line 3: Should be at least 20, to cover the larger range `BETWEEN '2024-01-17' AND '2024-02-06'` from cohort 2.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+end_date = '2024-04-22'
```
danielkberry, line 4: This should be unnecessary. Not saying you have to take it out, but should be superfluous.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+
+## Segments
+
+[segments]
+[segments.cohort_1]
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-01-30'"
+data_source = "segment_enrollments"
+
+[segments.cohort_2]
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-02-06'"
```
danielkberry, line 15: Was this one intended to be overlapping with cohort 1? or disjoint? Right now it overlaps

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+[segments.data_sources.segment_enrollments]
+client_id_column = "client_id"
+submission_date_column = "enrollment_date"
+friendly_name = "Enrolment table used to calculate above segments"
+from_expression = """(
+    SELECT
+        e.client_id,
+        `mozfun.map.get_key`(e.event_map_values, 'branch')
+            AS branch,
+        MIN(e.submission_date) AS enrollment_date,
```
danielkberry, line 27: We'll need to rename `enrollment_date` here because the actual enrollment date will show in the queries and now we have two columns named the same. How about something like `segment_enrollment_date`

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-01-30'"
+data_source = "segment_enrollments"
+
+[segments.cohort_2]
+select_expression = "enrollment_date BETWEEN '2024-01-17' AND '2024-02-06'"
+data_source = "segment_enrollments"
+
+[segments.data_sources.segment_enrollments]
+client_id_column = "client_id"
+submission_date_column = "enrollment_date"
```
danielkberry, line 20: Same as below, rename to `segment_enrollment_date`

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+      QUALIFY
+        COUNTIF(new_session_start AND NOT is_serp) OVER (
+          PARTITION BY glean_client_id, client_session_id
+        ) > 0
+    )
+    SELECT
+      *
+    FROM
+      urlbar_sessions
+)"""
```
danielkberry, line 694: Yikes, what a query. Definitely in the realm of "this should be implemented as a table instead of calculated on the fly by Jetstream". I'm concerned about the complexity and the runtime of this query. I'm not going to block on changing that here, we can try to run as is, but something to consider in the future.

danielkberry (changes requested):

jetstream/sponsored-suggestion-placement-phase-2.toml
dzeber: Totally agree, and I have a task to work on this as we start to use this approach more. This is based on a prototype query I used for the sponsored suggestion phase 1 experiment (also in the experiment config). FWIW it gave the desired results, but I'm not sure about performance implications. I'm working on a proposal so we can align on the methodology, and then we can implement in a table.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,694 @@
...
+      -- urlbar.exposure events sequence info
+      SELECT
+        client_info.client_id AS glean_client_id,
+        ping_info.seq,
+        event_timestamp,
+        TRUE AS exposed,
+      FROM
+        `mozdata.firefox_desktop.events_unnested`
+      WHERE
+        event_category = 'urlbar' AND event_name = 'exposure'
```
dzeber, line 528: We don't need to deal with exposures in this experiment, since all users that were "exposed" by typing a sponsored keyword actually saw a sponsored result, which is recorded in the urlbar events. Are stakeholders asking for an exposure-based analysis? If so, you can simplify these `urlbar*` CTEs to just adding a flag when there was a sponsored result impression.

dzeber (commented):
Looks good overall. It looks like many of the metrics are carried over from phase 1 - are stakeholders looking to carry out the same analyses? For exposures, you'd need to update the definition as discussed below.

jetstream/sponsored-suggestion-placement-phase-2.toml
sergiosonline: Hey, @dzeber . Agreed, but I'm staying consistent with the experiment brief, which does indicate exposure-based analysis. I'm happy to update this query as much as possible--so far I just copy-pasted your initial query since stakeholders are interested in the exact same set of metrics.

jetstream/sponsored-suggestion-placement-phase-2.toml
sergiosonline: Oh yes. I'll disconnect them.

One more thing, if I want to get a read on the entire population enrolled in the experiment, will that come automatically, or do I need to specify and additional cohort? If I do this, will `enrollment_period` be the maximum of enrollment periods across all cohorts?

danielkberry: That'll be automatic, it'll be the `all` segment. And yes, set `enrollment_period` to cover all the cohorts. So if you have 4 weekly cohorts, set `enrollment_period` to 28 and you'll get 5 segments (`all`, `cohort_1`, ...)

danielkberry (approved):
LGTM

jetstream/sponsored-suggestion-placement-phase-2.toml
dzeber: Sounds good - so yeah, rather than querying for `urlbar.exposure` events, you can change `urlbar_joined` to
```sql
select
  *,
  (select logical_or(r.product_result_type = 'admarketplace_sponsored') from unnest(results) as r) as exposed
from ...
```

sergiosonline: Thanks, @dzeber . I adopted your tip and simplified the workflow. Overall resulting counts are identical.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,658 @@
...
+
+[data_sources]
+
+[data_sources.tagged_google_serp_events]
+from_expression = """(
+    SELECT
+        * EXCEPT (is_ad_component),
+        component IN ('ad_carousel', 'ad_image_row', 'ad_link', 'ad_sidebar',
+        'ad_sitelink') AS ad_component
+    FROM `mozdata.firefox_desktop.serp_events`
```
dzeber, line 502:
```suggestion
    FROM `moz-fx-data-shared-prod.firefox_desktop_derived.serp_events_v1`
```
I suggest querying from the underlying v1 table just for this experiment as we're working on a v2, and this will ensure it'll use the expected table.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,658 @@
...
+        LOGICAL_OR(ad_component AND num_ads_loaded_reported > 0) AS has_ads_loaded,
+        LOGICAL_OR(ad_component AND num_ads_showing > 0 AND num_clicks > 0) AS has_ad_click,
+        SUM(IF(ad_component AND num_ads_showing > 0, num_clicks, 0)) AS num_ad_clicks,
+      FROM (
+        -- redefine ad componet to exclude non-monetized components
+        SELECT
+          * EXCEPT (is_ad_component),
+          component IN ('ad_carousel', 'ad_image_row', 'ad_link', 'ad_sidebar', 'ad_sitelink') AS ad_component
+        FROM
+          `mozdata.firefox_desktop.serp_events`
```
dzeber, line 563: Same as above, can update source table

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,658 @@
...
+        *
+      FROM
+        `mozdata.firefox_desktop.urlbar_events`
+      WHERE
+        is_terminal
+        AND submission_date BETWEEN '2024-01-16' AND '2024-04-22'
+        AND mozfun.map.get_key(experiments, 'sponsored-suggestion-placement-phase-2') IS NOT NULL
+    ),
+    urlbar_with_exposed AS (
+      -- interleave exposure events with main urlbar events
```
dzeber, line 530: Should remove these comments for clarity as we're not working with the exposure events here

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,658 @@
...
+        STRUCT(
+          exposed,
+          event_action,
+          product_selected_result,
+          (
+            SELECT
+              COUNTIF(product_result_type = 'admarketplace_sponsored') > 0
+            FROM
+              UNNEST(results)
+          ) AS has_sponsored_impression
```
dzeber, line 594: Noting that `has_sponsored_impression` should be the same as `exposed` for this experiment. Probably best to leave as is so as not to have to modify the metric definitions.

jetstream/sponsored-suggestion-placement-phase-2.toml
```diff
@@ -0,0 +1,658 @@
...
+description = "Per-client rate of SERP ad clicks per exposed urlbar search session"
+
+[metrics.exposed_ad_clicks_per_session.statistics.bootstrap_mean]
+[metrics.exposed_ad_clicks_per_session.statistics.deciles]
+
+[metrics.exposed_adm_impressions_per_session]
+select_expression = """SAFE_DIVIDE(
+    COUNTIF(exposed_session AND urlbar_fields.has_sponsored_impression),
+    COUNTIF(exposed_session AND new_session_start)
+)"""
```
dzeber, line 470: FWIW looks like this metric will end up being 1 per the definition of exposure.

dzeber (approved):

PR #405, opened 2024-05-16
Update all metrics on ios health to use ratios

OrlaM:
Updated the cpu exception and large file write metrics to use ratios instead.

Removed the total baseline pings and dirty_startup sum visualizations, there should just be 3 charts on the dashboard now. Please let me know if that is not the correct way to achieve that change.

### DIFF
--- a/opmon/firefox-ios-health.toml
+++ b/opmon/firefox-ios-health.toml
@@ -12,6 +12,8 @@ data_source = "baseline_v2"
 monitor_entire_population = true
 channel = "release"
 
+[metrics.total_baseline_pings.statistics.sum]
+
 [metrics.large_file_write.statistics.total_ratio]
 denominator_metric = "total_baseline_pings"
 [metrics.cpu_exception.statistics.total_ratio]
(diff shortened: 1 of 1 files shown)


opmon/firefox-ios-health.toml
```diff
@@ -12,14 +12,10 @@ data_source = "baseline_v2"
...
 channel = "release"
 
-[metrics.dirty_startup.statistics]
-sum = {}
-[metrics.large_file_write.statistics]
-sum = {}
-[metrics.cpu_exception.statistics]
-sum = {}
-[metrics.total_baseline_pings.statistics.sum]
-
```
scholtzan, line 22:
```suggestion
[metrics.total_baseline_pings.statistics.sum]
```

opmon/firefox-ios-health.toml
```diff
@@ -12,14 +12,10 @@ data_source = "baseline_v2"
...
+[metrics.large_file_write.statistics.total_ratio]
```
scholtzan, line 15:
```suggestion
[metrics.total_baseline_pings.statistics.sum]

[metrics.large_file_write.statistics.total_ratio]
```

scholtzan (approved):

PR #433, opened 2024-05-30
Add dimensions to ios health dashboard

OrlaM:
Add channel dimension to the iOS beta dashboard.


opmon/firefox-ios-beta-health.toml
```diff
@@ -10,15 +10,20 @@ metrics = ['dirty_startup', 'large_file_write', 'hang_exception', 'cpu_exception
...
+dimensions = ["release", "channel"]
 
 [dimensions.release]
 data_source = "baseline_v2"
 select_expression = "client_info.app_display_version"
 friendly_name = "Release Version"
 description = "Release version"
 
+[dimensions.channel]
+data_source = "main"
```
scholtzan, line 22:
```suggestion
data_source = "baseline_v2"
```

`main` is only for FF desktop telemetry data, but `baseline` has channel information too

scholtzan (approved):

opmon/firefox-ios-beta-health.toml
OrlaM: Copy paste error, good catch 🙏

PR #444, opened 2024-06-07
Add metrics for SERP events

dzeber:


jaredsnyder (approved):
Discussed what we wanted to include and this matches what we decided.  Looks good!

PR #447, opened 2024-06-10
Update yelp-suggestions-pilot-v3.toml

pissac17:
adding users counts that disabled fx suggest and sponsored suggest


dzeber (commented):
As discussed, the metrics as written would behave like boolean proportions. However, we maybe don't need to add these to the config.

PR #452, opened 2024-06-11
Add boolean Pocket click metric

m-d-bowerman:
Adds metrics to test for changes in share of clients that have any Pocket clicks.


sergiosonline (approved):
looks good

PR #473, opened 2024-06-20
Create weather-suggestions-simplified.toml

teonbrooks:
This experiment config file is to provide additional metrics for the simplified weather experiment


m-d-bowerman (approved):
The config file itself looks good, and the `select_expression`s of the metrics look consistent with other definitions I see that use `urlbar_events_unnested`. Just double those expressions for correctness, though, as I'm not familiar with the data source.

teonbrooks:
@m-d-bowerman, thanks for the review!

PR #476, opened 2024-06-21
Update weather-suggestions-simplified.toml

teonbrooks:
fixing the data source for all the non-impression metrics


dzeber (approved):
LGTM

PR #486, opened 2024-06-26
[MRG] Create fakespot-in-suggest.toml

teonbrooks:
First draft of the metrics needed for the fakespot experiment.
- Need to confirm `result_type`
- Need to check if there are any other metrics


jetstream/fakespot-in-suggest.toml
```diff
@@ -0,0 +1,203 @@
...
+           "fakespot_bestbuy_ctr",
+           "fakespot_walmart_ctr",
+           "fakespot_amazon_not_relevant_dismissals",
+           "fakespot_bestbuy_not_relevant_dismissals",
+           "fakespot_walmart_not_relevant_dismissals",
+           "fx_urlbar_quick_suggest_remote_settings_latency_ms"]
+
+## Impressions
+[metrics.admarketplace_sponsored_impressions]
+select_expression = "COUNT(DISTINCT IF(is_terminal AND result.result_type like '%adm_sponsored', event_id, NULL))"
```
rebecca-burwei, line 40: Is there a way to re-use this UDF? https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/result_type_to_product_name/udf.sql

jetstream/fakespot-in-suggest.toml
```diff
@@ -0,0 +1,203 @@
...
+data_source = "urlbar_events_unnested_results"
+[metrics.fakespot_bestbuy_impressions.statistics.bootstrap_mean]
+
+[metrics.fakespot_walmart_impressions]
+select_expression = "COUNT(DISTINCT IF(is_terminal AND result.result_type = 'rust_fakespot_walmart', event_id, NULL))"
+data_source = "urlbar_events_unnested_results"
+[metrics.fakespot_walmart_impressions.statistics.bootstrap_mean]
+
+[metrics.fakespot_impressions]
+select_expression = "COUNT(DISTINCT IF(is_terminal AND result.result_type in ('rust_fakespot_amazon', 'rust_fakespot_bestbuy', 'rust_fakespot_walmart'), event_id, NULL))"
```
rebecca-burwei, line 60: same here. I'd like to keep the business logic in that UDF if possible, so it's not scattered across multiple repos. https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/norm/result_type_to_product_name/udf.sql

rebecca-burwei (changes requested):
Could you give the UDF a shot? It's OK if you investigate for 15 min and decide it's too big a lift. Just, especially since you're handing off this work, it would be nice if the business logic were in the UDF with the rest of them.

jetstream/fakespot-in-suggest.toml
teonbrooks: for the impressions, we don't currently map the product name to the results array, only to the selected result. this poses a problem when you want to calculate the impression since these have the original name and not the remapped values. 
I am not aware of a way to use the udf here to keep the business logic in one place and keep this DRY

jetstream/fakespot-in-suggest.toml
teonbrooks: since this is a new product type for the experiment/pilot, we haven't brought into the udf mentioned. I think that would be useful thing to do if we decide to move forward with the feature but as of now, this would be the best way to calculate the impression

teonbrooks:
@rebecca-burwei, thanks for the review. I updated the admarketplace to simplify and keep the business logic in the udf. I added a response to your followup comment on the fakespot-related suggestion. unfortunately, I think that it would require more work to add it to the udf given that it is an experimental feature so I think it makes sense to leave it in the experiment config until the feature has graduated (if that's the conclusion from the experiment)

rebecca-burwei (approved):
Thanks! looks good

rebecca-burwei:
I think the validate-jetstream error is because there are 2 metrics called fakespot_amazon_ctr in lines 75 and 138

teonbrooks:
good catch. ended up being an error in all the impression rate definitions. it was a copy-paste error

PR #490, opened 2024-06-28
Redo Generated Looker explore for PPA measurements

curtismorales:


braunreyes:
@scholtzan can you advise on the validate-looker error?

scholtzan:
The service account used for dry running doesn't seem to have access to the table. I'll push an update to use a different dry run endpoint.
But it's good to go now that the table exists

PR #509, opened 2024-07-09
[ci skip-rerun] re-enable covariate adjustment

danielkberry:
Reverts mozilla/metric-hub#498. 

When we previously pushed this, we encountered some errors. Those were largely fixed in https://github.com/mozilla/mozanalysis/pull/266. However, when an experiment is rerun, there is now a race condition between the analysis periods. Later analysis periods (e.g., week & overall) will depend on prior analysis periods (i.e., preenrollment_week) which may or may not have completed. In the case where covariate adjustment using preenrollment_week was specified but preenrollment_week is not available, we will now fall back (https://github.com/mozilla/mozanalysis/pull/268) to unadjusted inferences and emit a warning.  Since these improvements are now in Jetstream (https://github.com/mozilla/jetstream/pull/2085), we can attempt to reenable the inferences. 

Once the Jetstream rearchitecture is complete, we expect to be able to specify dependencies between periods and avoid the need to fallback.

### DIFF
--- a/jetstream/defaults/fenix.toml
+++ b/jetstream/defaults/fenix.toml
@@ -59,12 +59,21 @@ binomial = {}
 deciles = {}
 bootstrap_mean = { drop_highest = 0 }
 empirical_cdf = {}
+[metrics.days_of_use.statistics.linear_model_mean]
+drop_highest = 0.0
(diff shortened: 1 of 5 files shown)


mikewilli (approved):
2nd time's the charm!

PR #513, opened 2024-07-10
added pdf handler metrics and outcome

betling:
added metrics for when firefox was invoked or launched and created a new outcome with those metrics.  this work completes the second request in this [ticket](https://mozilla-hub.atlassian.net/browse/DS-3669).


danielkberry (approved):
LGTM

PR #526, opened 2024-07-18
update client-level DAU definitions

bochocki:
Adds client-level DAU definitions for each product as requested in [DS-3752](https://mozilla-hub.atlassian.net/browse/DS-3752).

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -728,3 +728,4 @@ from_expression = """(
 )"""
 submission_date_column = "submission_date"
 client_id_column = "client_id"
+deprecated = false
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -97,6 +97,21 @@ description = """
     For questions please contact bochocki@mozilla.com or firefox-kpi@mozilla.com.
 """
 owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
+deprecated = true
+
+[metrics.client_level_daily_active_users_v3]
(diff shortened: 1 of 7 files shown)


bani (approved):

jmsilverman (approved):
I know there's a way to run Acryl/Datahub locally to check how the changes will actually render so I'm curious if you've done that, but just reviewing the code everything LGTM!

I guess one NIT is how consistent we want to be using `deprecated=false` (which is the default, of course). I'm not sure what best practice is here, but I think it's fine as is?

bochocki:
> Thanks for adding these new definitions. Are we going to make a separate PR to update the experiment guardrails to point to these?

That sounds reasonable to me. I'm not familiar with how the experiment guardrails are implemented and wouldn't know how to update them. The concerns are separate enough that having two PRs seems sensible.

bochocki:
> I guess one NIT is how consistent we want to be using deprecated=false (which is the default, of course). I'm not sure what best practice is here, but I think it's fine as is?

I'm not sure I follow. I _think_ I've set deprecated = false for all the metrics and updated deprecated = true where applicable, but please lmk if I'm missing something.

Maybe you're noting that the data_sources don't have a deprecation flag; I didn't use the field there because I assumed it wasn't a valid field on the data_sources: It's not described in the docs ([here, I think](https://docs.telemetry.mozilla.org/concepts/metric_hub#data_sources-section)) and I didn't see it used on another data_source. If it _can_ be used it does seem useful to include and I'd be happy to add it. I'll check in #data-help.

PR #523, opened 2024-07-18
Add OKR metrics to Desktop and Mobile products

bochocki:
Adds OKR definitions (Engagement Rate, Retention Rate, New Profile Retention Rate) to Metric Hub ([DS-3762](https://mozilla-hub.atlassian.net/browse/DS-3762)).

PR #483, opened 2024-06-25
Clarify source naming; update link to generated SQL

fbertsch:
We were using the source name `active_users_aggregates_v1`, which actually pointed to a view. This now clarifies that we're pointing to the view, and filtering to just desktop.

Also updated the link in the metric definition to point to the generated source query.

PR #527, opened 2024-07-18
add a sample script to validate Metric Hub definitions

bochocki:
I used a script similar to this to validate definition changes in https://github.com/mozilla/metric-hub/pull/483 and https://github.com/mozilla/metric-hub/pull/523 after merging. @jmsilverman suggested it could be helpful to check the script into the repo.

Open to a discussion about whether or not this is useful and, if not, if there's something we can do instead.


definitions/README.md
```diff
@@ -2,3 +2,84 @@
...
+    # Modify the metric source table string so that it formats nicely in the query.
+    from_expression = metric.data_source._from_expr.replace("\n", "\n" + " " * 15)
+
+    query = dedent(
+        f"""
+        SELECT {submission_date_column},
+               {metric.select_expr} AS value
+          FROM {from_expression}
+         WHERE {submission_date_column} BETWEEN '{start_date}' AND '{end_date}'
+         GROUP BY {submission_date_column}
```
jmsilverman, line 43: I'd suggesting adding `         ORDER BY {submission_date_column}` under here so that the results are always in chronological order.

definitions/README.md
```diff
@@ -2,3 +2,84 @@
...
+
+    # ensure submission_date has type 'date'
+    df[submission_date_column] = pd.to_datetime(df[submission_date_column]).dt.date
+
+    return df
+```
+
+Third, verify that the relevant definitions can be queried.
+
+```python3
```
jmsilverman, line 64: I personally like factoring out the hardcoded variables from below up here so they're more obvious to update as needed, i.e.:

```
# BQ project
bq_project = "moz-fx-data-bq-data-science"

# start & end dates to query metrics
start_date = "2023-12-20"
end_date = "2023-12-17"
```

(and replace the hardcoded values below with the variable names above).

jmsilverman (changes requested):
Requesting a couple small code changes via inline comments.

As Brad mentioned, I think this script is useful to have in the repo and I ran it in Colab and it worked great!

I'm not totally sure if it belongs in the README vs. as a standalone script/file, so I'm curious to hear other people's opinions on that.

definitions/README.md
```diff
@@ -2,3 +2,84 @@
...
+    end_date = pd.to_datetime(end_date).date()
+
+    # Set useful attributes based on the Metric Hub definition
+    metric = ConfigLoader.get_metric(metric_slug=metric_slug, app_name=app_name)
+    submission_date_column = metric.data_source.submission_date_column
+
+    # Modify the metric source table string so that it formats nicely in the query.
+    from_expression = metric.data_source._from_expr.replace("\n", "\n" + " " * 15)
+
+    query = dedent(
```
danielkberry, line 37: What about adding the query builder as a method inside of `DataSource`, something like `DataSource.build_validation_query(start_date: str, end_date: str) -> str`. Then we can expose that through the `Metric` through something like: 

```python3
# in Metric class
def build_validation_query(self, start_date: str, end_date: str) -> str:
    return self.data_source.build_validation_query(start_date, end_date)
```

In the `DataSource` method, we can handle modifying the from expression like how `build_query` [does it](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/metrics.py#L177). Additionally, that gives us a separate method to unit test. Right now, the logic is hidden within this larger function.

definitions/README.md
```diff
@@ -2,3 +2,84 @@
...
+import pandas as pd
+
+from google.cloud import bigquery
+from google.colab import auth
+from mozanalysis.config import ConfigLoader
+from textwrap import dedent
+
+auth.authenticate_user()
+
+def fetch_metric_hub_definition(start_date, end_date, metric_slug, app_name, bq_project):
```
danielkberry, line 26: What do you think about a name like `fetch_validation_dataframe`? When I think of a metric "definition" I think of metadata, whereas this function returns materialized data.

definitions/README.md
danielkberry: If we do that, this function can be simplified to: 

```python3
metric = ConfigLoader.get_metric(metric_slug=metric_slug, app_name=app_name)
query = metric.build_validation_query(start_date, end_date)
df = bigquery.Client(project=bq_project).query(query).to_dataframe()
df[submission_date_column] = pd.to_datetime(df[submission_date_column]).dt.date
return df
```

definitions/README.md
danielkberry, line 26: Also, can we add type annotations to the parameters?

definitions/README.md
danielkberry, line 43: Most metric definitions assume that we're grouping by `client_id` too, so we should add that to the `GROUP BY`

definitions/README.md
```diff
@@ -2,3 +2,84 @@
...
+for app_name, metric_slugs in definitions.items():
+    for metric_slug in metric_slugs:
+        df = fetch_metric_hub_definition(
+            start_date="2024-03-14",
+            end_date="2024-03-21",
+            metric_slug=metric_slug,
+            app_name=app_name,
+            bq_project="moz-fx-data-bq-data-science"
+        )
+        display(df.set_index(df.columns[0]).transpose())
```
danielkberry, line 84: Grabbing the zeroth column seems fragile. We already have the column name (`metric_slug`) here, why don't we select that one explicitly.

danielkberry (changes requested):
I think this is a great idea, just a few suggestions

definitions/README.md
bochocki: It is definitely fragile. I'm pivoting on the date column mostly so that I can get a display that doesn't take up too much space and looks nice:

image

It could be more robust to update the first line of the query to

```
SELECT {submission_date_column} AS submission_date,
```
and then `df.set_index("submission_date")` should always work as expected.

definitions/README.md
bochocki: I really, really like the idea of having an attribute on Metric that provides a query. I don't know if it's difficult to do in the general case, but it's something I think could be useful outside of this case. For example, there's a [class in the forecasting code](https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting/kpi_forecasting/metric_hub.py) whose main goal is to build a query from Metric Hub components.

definitions/README.md
bochocki: Is there a programmatic way to figure out what other groupby dimensions might be?

bochocki:
Changed to WIP so it's clear this isn't meant to be merged right now. I also addressed some of the comments.

I think that having a way to build queries from Metric Hub objects would be a useful addition to `mozanalysis` and would probably be more valuable than documenting a script like this somewhere.

PR #536, opened 2024-07-26
Update fakespot-in-suggest.toml

scholtzan:
The experiment hasn't been launched yet. It seems like the required CI checks weren't required anymore so this PR got merged (even though it shouldn't).

Adding `skip = true` until the experiment is actually being launched to prevent jetstream errors from being raised

cc @teonbrooks


teonbrooks:
ah, cool, sorry didn't know that. thanks for the fix!

PR #553, opened 2024-08-07
feat: add aggregation_units to metric and datasource

mikewilli:
Step 1 towards supporting profile management feature. 

- adds support for `aggregation_units` to metrics and data_sources, and defaults both to what is currently supported, `client_id` only
- ensures that the data_source supports (at least) all aggregations that the metric supports


scholtzan:
I wonder if calling it `group_by` would be a better name?

mikewilli:
To me `group_by` implies that we would group by the configured value(s). In this case, it's only meant to indicate which fields are supported as a `group_by`. Might still be fine / better than `aggregation_units`, but just concerned that people will think we are grouping by everything in the list. `group_by_support` maybe?

I'm definitely open to a different name though, and @danielkberry and I discussed it a bit already (`analysis_units` was another one we discussed).

lib/metric-config-parser/metric_config_parser/tests/test_analysis.py
```diff
@@ -763,3 +765,68 @@ def test_multiple_nesting(self, experiments, config_collection):
...
+            [data_sources.eggs]
+            from_expression = "england.camelot"
+            client_id_column = "client_info.client_id"
+            {ds_units}
+            """
+        )
+
+        spec = AnalysisSpec.from_dict(toml.loads(config_str))
+        # print(spec.metrics.weekly[0].resolve(spec, , config_collection))
+        print(f"metric: {metric_units}, ds: {ds_units}")
```
danielkberry, line 829: Are these leftover from debugging?

lib/metric-config-parser/metric_config_parser/tests/test_analysis.py
```diff
@@ -763,3 +765,68 @@ def test_multiple_nesting(self, experiments, config_collection):
...
+
+        metric = [m for m in cfg.metrics[AnalysisPeriod.WEEK] if m.metric.name == "active_hours"][
+            0
+        ].metric
+        assert metric.aggregation_units == [AnalysisUnit.CLIENT]
+
+        data_source = metric.data_source
+        assert data_source.aggregation_units == [AnalysisUnit.CLIENT]
+
+    @pytest.mark.parametrize(
```
danielkberry, line 790: Excellent use of parametrize here. This test checks all the invalid combinations, would it be useful or unneccessary to have a similar test to test the valid combinations?

lib/metric-config-parser/metric_config_parser/metric.py
```diff
@@ -239,6 +242,7 @@ def resolve(
                     owner=[self.owner] if isinstance(self.owner, str) else self.owner,
                     deprecated=self.deprecated,
                     level=self.level,
+                    aggregation_units=self.aggregation_units or [AnalysisUnit.CLIENT],
```
danielkberry, line 245: We've got a few of these `self.parameter or [AnalysisUnit.CLIENT]` sprinkled around. What do you think about using attr's validators and converters instead (like [here](https://github.com/mozilla/mozanalysis/pull/273/files#diff-d3f3c2d9005e084f57601bb9527d9af10b834c513b2faf2dc5e926273b7b01e1R97-R102)). This would allow `Metric.__init__` to take a `aggregation_units=None` and convert that internally to `self.aggregation_units=[AnalysisUnit.CLIENT]`. It would move the default behavior logic internal to the class, instead of needing to be handled by callers.

lib/metric-config-parser/metric_config_parser/metric.py
```diff
@@ -247,16 +251,31 @@ def resolve(
                 ]
                 metric_definition.statistics = self.statistics
                 metric_summary = metric_definition.resolve(spec, conf, configs)
+                metric_definition.aggregation_units = self.aggregation_units or [
+                    AnalysisUnit.CLIENT
+                ]
```
danielkberry, line 256: What's the purpose of this `elif metric_definition` block here? A metric definition is missing a select expression and a data source, so we pull the rest of the definition from the config collection and overwrite its aggregation units? I think I'm a bit lost.

lib/metric-config-parser/metric_config_parser/data_source.py
```diff
@@ -77,6 +78,16 @@ class DataSource:
             `{dataset}` in from_expr if a value is not provided
             at runtime. Mandatory if from_expr contains a
             `{dataset}` parameter.
+        build_id_column (str, optional):
+            Default 'SAFE.SUBSTR(application.build_id, 0, 8)'.
+        friendly_name (str, optional)
+        description (str, optional)
+        joins (list[DataSourceJoin], optional)
+        columns_as_dimensions (bool, optional): Default false.
+        aggregation_units (list[AnalysisUnit], optional): denotes which
```
danielkberry, line 87: I recall our discussion about `aggregation_unit` vs `analysis_unit`, but I'd assumed we'd be consistent: either `aggregation_unit/AggregationUnit` or `analysis_unit/AnalysisUnit`. Are we mixing names by calling the parameter `aggregation_unit` but the type `AnalysisUnit`?

lib/metric-config-parser/metric_config_parser/data_source.py
```diff
@@ -90,6 +101,7 @@ class DataSource:
     description = attr.ib(default=None, type=str)
     joins = attr.ib(default=None, type=List[DataSourceJoin])
     columns_as_dimensions = attr.ib(default=False, type=bool)
+    aggregation_units = attr.ib(default=[AnalysisUnit.CLIENT], type=List[AnalysisUnit])
```
danielkberry, line 104: Without a validator, is this guaranteed to be of type list? The doc string says its optional, but are data sources ever initialized with `aggregation_units = None`? If that's the case, at runtime the type of this parameter could be `None` despite the type annotation saying `list[AnalysisUnit]`.

lib/metric-config-parser/metric_config_parser/data_source.py
```diff
@@ -190,6 +206,7 @@ def resolve(
             "friendly_name",
             "description",
             "columns_as_dimensions",
+            "aggregation_units",
```
danielkberry, line 209: I _think_ this guards against my comment above, but is there a way we can be sure? If someone writes a config and omits the parameter, then here nothing will be set in the `params` dict, so nothing will be passed to `DataSource.__init__`, so it should use the `[AnalysisUnit.CLIENT]` default then... I think...

lib/metric-config-parser/metric_config_parser/tests/test_config.py
```diff
@@ -505,6 +507,92 @@ def test_invalid_metric_level(self):
...
+            data_source = "baseline"
+            {metric_units}
+
+            [data_sources.baseline]
+            from_expression = "mozdata.search.baseline"
+            experiments_column_type = "simple"
+            {ds_units}
+            """
+        )
+        with pytest.raises(Exception):
```
danielkberry, line 588: Is there any way to make this error more specific? `Exception` seems vague

danielkberry (changes requested):
Looks great! A few small suggestions and questions. I'm excited to incorporate this into my mozanalysis PR.

lib/metric-config-parser/metric_config_parser/tests/test_config.py
mikewilli: Yea I didn't like this either. The exception is a `ValueError`, but it's actually multiple `ValueError`s within a group of exceptions. For example, because it expects a list, it parses a string value as a list of characters, none of which are valid `AnalysisUnit`s, and each of those characters produces a `ValueError`, raised together in a group. In Python 3.11 there was an `ExceptionGroup` introduced, which might give us more options, but I opted for the overly broad solution here in line with other parts of the code.;

lib/metric-config-parser/metric_config_parser/tests/test_analysis.py
mikewilli: Yep!

lib/metric-config-parser/metric_config_parser/metric.py
mikewilli: Yea the idea of this code is that if the current metric is missing a `select_expression` or `data_source`, it is probably augmenting/overriding a metric defined somewhere else. We get that definition with `metric_definition = configs.get_metric_definition(self.name, conf.app_name)` and then override a few fields.

What I'm not clear on is why we only override a few select fields here and not just take every field that is defined. @scholtzan any thoughts here?

lib/metric-config-parser/metric_config_parser/metric.py
mikewilli: I actually started doing this on the DataSource side where we do use validators for some fields. But then I ended up deleting it because defining the type accomplished the same thing and there wasn't any additional validation needed. 

What's the `converter` do that `default` doesn't?

Also, if we do want to go down this road, I'd suggest we consider it separately and aim to cover more than just the new `aggregation_units` field.

danielkberry: +1 to "let's do it later, not necessary for this PR"

As I understand it, `default` defines the default argument value while `converter` applies a function to the argument. Converters allow you to differentiate between the types of the parameters and the types of the attributes

So
```python
class MyClass:
  x = attr.field(default = None, type = str | None)
```
will define a class with a constructor like: 
```python
MyClass.__init__(self, x: str | None = None):
  self.x = x # type will be str | None
```

If you add a converter: 
```python
class MyClass: 
  x = attr.field(default = None, type = str, converter=attr.converter.default_if_none("default"))
```
you'll get a class with a constructor like: 
```python
MyClass.__init__(self, x: str | None = None):
  self.x = x if x is not None else "default" # type will be str, no longer optional
``` 

I think it comes down to a question of 1) what parameters do we want callers to be able to pass to the constructor and 2) what type do we want the class attributes to be.

lib/metric-config-parser/metric_config_parser/tests/test_analysis.py
mikewilli: Sure, can't hurt.

lib/metric-config-parser/metric_config_parser/data_source.py
mikewilli: Hm.. I think that should be disallowed. And I don't think you can pass the literal python `None` as a value, it would have to be a string, which would raise a `ValueError` the same way any other string does.

lib/metric-config-parser/metric_config_parser/metric.py
scholtzan: Yeah, this is to override a metric defined somewhere else. I think these few select fields are the ones relevant for analysis. Might make sense to also override the other available fields

lib/metric-config-parser/metric_config_parser/data_source.py
scholtzan: Right, toml doesn't support null/None. If this field wouldn't have a default value and would not be explicitly set, then that's when it would be `None`

lib/metric-config-parser/metric_config_parser/metric.py
mikewilli: Got it, thanks for the examples.

I'm in favor of a bit more rigorously defined set of schemas for the various metric-hub entities. I was originally picturing pydantic schemas in nimbus-schemas package, but given that these aren't Nimbus specific that's not quite the right place for them. This might be a better approach (and a better location).

danielkberry: I'm not that familiar with Pydantic. Do you think it's worth migrating to that as the solution?

mikewilli: https://github.com/mozilla/metric-hub/issues/554

lib/metric-config-parser/metric_config_parser/data_source.py
mikewilli: In my head there's sort of a distinction between aggregation unit (generic, what can we group by) and the analysis unit (specific to experimentation), but you're probably right. I'll go ahead with changing everything to `analysis_units`.

lib/metric-config-parser/metric_config_parser/metric.py
mikewilli: Hm... I like pydantic, but my first thought is that it's not worth migrating. It was a bit of a pain incorporating pydantic stuff into jetstream for nimbus-schemas though partly due to the mix of attrs and pydantic, so having everything consistent would be nice. Seems like an ADR would be appropriate, though I'm not exactly itching to write it at the moment.

lib/metric-config-parser/metric_config_parser/metric.py
danielkberry: Should we gate this on `self.aggregation_units is not None`? What if the metric elsewhere (`metric_definition`) explicitly specifies `profile_group_id` as the `aggregation_unit` and the overriding metric does not specify anything? In that case, this will override the profile unit with a client unit, right? Is that intended?

lib/metric-config-parser/metric_config_parser/metric.py
danielkberry: Agree. Let's plan to circle back to this later, shouldn't block this PR.

danielkberry:
There's just one open question, then I can re-review.

lib/metric-config-parser/metric_config_parser/metric.py
mikewilli: Yea good call. I added a test for this to verify the behavior you described, and fixed it by adding another option to the `or` chain. Now we'll use (in order) the `self` value if available, or the `metric_definition` if available, or the default if neither is defined.

Of note, this was the same for `analysis_bases` so fixed that here too.

danielkberry (approved):
Looks like all the changes are made, thanks! LGTM

PR #560, opened 2024-08-19
Update fakespot-in-suggest.toml

pissac17:
adding new grades and rating metrics as defines here: https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/urlbar_fakespot_engagement

### DIFF
--- a/jetstream/fakespot-in-suggest.toml
+++ b/jetstream/fakespot-in-suggest.toml
@@ -6,18 +6,23 @@ daily = ["admarketplace_sponsored_impressions",
           "fakespot_amazon_impressions",
           "fakespot_bestbuy_impressions",
           "fakespot_walmart_impressions",
+          "fakespot_impressions",
           "fakespot_amazon_impression_rate",
           "fakespot_bestbuy_impression_rate",
           "fakespot_walmart_impression_rate",
+          "fakespot_impression_rate",
           "fakespot_amazon_clicks",
           "fakespot_bestbuy_clicks",
           "fakespot_walmart_clicks",
+          "fakespot_clicks",
           "fakespot_amazon_ctr",
           "fakespot_bestbuy_ctr",
           "fakespot_walmart_ctr",
+          "fakespot_ctr",
           "fakespot_amazon_not_relevant_dismissals",
           "fakespot_bestbuy_not_relevant_dismissals",
           "fakespot_walmart_not_relevant_dismissals",
+          "fakespot_not_relevant_dismissals",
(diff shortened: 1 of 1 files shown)


jetstream/fakespot-in-suggest.toml
```diff
@@ -199,12 +229,55 @@ select_expression = "COUNTIF(event_action = 'annoyance' and engaged_result_type
...
 [metrics.fx_urlbar_quick_suggest_remote_settings_latency_ms]
 [metrics.fx_urlbar_quick_suggest_remote_settings_latency_ms.statistics.bootstrap_mean]
+
+[data_sources]
+
+[data_sources.fakespot_events]
+from_expression = """(
+    SELECT 
+        metrics.uuid.legacy_telemetry_client_id AS client_id,
+        DATE(submission_timestamp) AS submission_date,
```
dzeber, line 263: I think the dataset definition needs to include the experiments column. Alternatively you can use `select *` rather than specifying specific columns, see an example [here](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L2014).

jetstream/fakespot-in-suggest.toml
```diff
@@ -199,12 +229,55 @@ select_expression = "COUNTIF(event_action = 'annoyance' and engaged_result_type
...
 [metrics.fakespot_not_relevant_dismissals.statistics.bootstrap_mean]
 
-## Review Checker
-[metrics.review_checker_displayed_count]
-select_expression = "COUNTIF(event_category = 'shopping' and event.name = 'surface_displayed')"
-data_source = "events"
-[metrics.review_checker_displayed_count.statistics.bootstrap_mean]
+
+[metrics.fakespot_4_star_A_grade]
+select_expression = "COUNTIF(mozfun.map.get_key(event_extra, 'grade') = 'A' and (CAST(mozfun.map.get_key(event_extra, 'rating') AS FLOAT64) >= 4 and CAST(mozfun.map.get_key(event_extra, 'rating') AS FLOAT64) < 5))"
```
dzeber, line 234: I recommend moving the logic for `grade` and `rating` (pulling from the map, casting) to the dataset query so that it doesn't need to be repeated each time.

dzeber (commented):
Looks good, I left a couple of comments inline.

If stakeholders are going to be looking at the Experimenter page, I recommend adding friendly_name and description for each metric.

For the CTR metrics the denominator is all sessions rather than engaged sessions - the descriptions need to be updated (I can't comment inline on that part).

PR #572, opened 2024-08-30
feat: add support for randomization_unit and analysis_unit in experiment config

mikewilli:
- adds a `randomization_unit` property to the experiment config to make it easier to access
- adds a `analysis_unit` property to the experiment config that is derived from `randomization_unit` for use by jetstream

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/experiment.py
+++ b/lib/metric-config-parser/metric_config_parser/experiment.py
@@ -5,11 +5,13 @@ from typing import TYPE_CHECKING, Any, List, Optional
 import attr
 import jinja2
 from jinja2 import StrictUndefined
+from mozilla_nimbus_schemas import RandomizationUnit
 
 if TYPE_CHECKING:
     from .config import ConfigCollection
     from .analysis import AnalysisSpec
 
+from . import AnalysisUnit
 from .errors import NoEndDateException, NoStartDateException
 from .exposure_signal import ExposureSignal, ExposureSignalDefinition
 from .segment import Segment, SegmentReference
@@ -149,6 +151,24 @@ class ExperimentConfiguration:
             return self.experiment.bucket_config.start
 
         return None
+
+    @property
+    def randomization_unit(self) -> Optional[RandomizationUnit]:
+        if hasattr(self.experiment, "bucket_config") and self.experiment.bucket_config is not None:
+            # this will raise a ValueError if the provided randomization_unit is invalid
+            return RandomizationUnit(self.experiment.bucket_config.randomization_unit)
+
+        return None
+
+    @property
+    def analysis_unit(self) -> Optional[AnalysisUnit]:
+        """Retrieve the appropriate analysis unit, which is
+        derived from the experiment's randomization unit.
+        """
+        if self.randomization_unit and self.randomization_unit == RandomizationUnit.GROUP_ID:
+            return AnalysisUnit.PROFILE_GROUP
+
+        return AnalysisUnit.CLIENT
 
     @property
     def enrollment_period(self) -> int:
--- a/lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
+++ b/lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
@@ -6,7 +6,9 @@ import pytest
 import pytz
 import toml
 from cattrs.errors import ClassValidationError
+from mozilla_nimbus_schemas import RandomizationUnit
 
+from metric_config_parser import AnalysisUnit
 from metric_config_parser.analysis import AnalysisSpec
 from metric_config_parser.errors import NoEndDateException
 from metric_config_parser.metric import AnalysisPeriod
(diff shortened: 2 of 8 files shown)


lib/metric-config-parser/metric_config_parser/experiment.py
```diff
@@ -150,6 +152,24 @@ def bucket_start(self) -> Optional[int]:
...
+            return RandomizationUnit(self.experiment.bucket_config.randomization_unit)
+
+        return None
+
+    @property
+    def analysis_unit(self) -> Optional[AnalysisUnit]:
+        """Retrieve the appropriate analysis unit, which is
+        derived from the experiment's randomization unit.
+        """
+        if self.randomization_unit and self.randomization_unit == RandomizationUnit.GROUP_ID:
```
danielkberry, line 168: Will this line also raise a ValueError if the randomization_unit is invalid?

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
```diff
@@ -376,6 +378,30 @@ def test_sample_size_none_experiment(self, experiments, config_collection):
...
+
+    def test_analysis_unit_configured(self, experiments, config_collection):
+        conf = dedent(
+            """
+            [experiment]
+            enrollment_period = 7
+            """
+        )
+        spec = AnalysisSpec.from_dict(toml.loads(conf))
+        cfg = spec.resolve(experiments[0], config_collection)
```
danielkberry, line 401: `experiments` is the list of `Experiment`s defined in `conftest.py`?

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
```diff
@@ -376,6 +378,30 @@ def test_sample_size_none_experiment(self, experiments, config_collection):
...
+        conf = dedent(
+            """
+            [experiment]
+            enrollment_period = 7
+            """
+        )
+        spec = AnalysisSpec.from_dict(toml.loads(conf))
+        cfg = spec.resolve(experiments[1], config_collection)
+        assert cfg.experiment.randomization_unit is None
+        assert cfg.experiment.analysis_unit == AnalysisUnit.CLIENT
```
danielkberry, line 391: For comprehensiveness, should we add a test that needs to parse the randomization unit into "normandy" or is that unnecessary?

danielkberry (approved):
Looks great! A few minor questions, but I wouldn't say you have to block on them.

lib/metric-config-parser/metric_config_parser/experiment.py
mikewilli: Yea correct. It'll call the function above to get the randomization unit and so will raise the ValueError.

mikewilli: I'll tack on that I thought this was acceptable because the randomization unit really should either be a) undefined (`None`) or b) valid. If it exists and is not a valid value that is bad. One thing I considered was if I should return `None` if the bucket_config is not None but randomization_unit is:
```py
    @property
    def randomization_unit(self) -> Optional[RandomizationUnit]:
        if hasattr(self.experiment, "bucket_config") and self.experiment.bucket_config is not None:
            ########## would add this ##########
            if self.experiment.bucket_config.randomization_unit is None:
                return None
            ########## until here ##########
            # this will raise a ValueError if the provided randomization_unit is invalid
            return RandomizationUnit(self.experiment.bucket_config.randomization_unit)

        return None
```

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
mikewilli: Yep, correct.

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
mikewilli: Yea, not a bad idea. Maybe I should rewrite the test as a parametrized thing instead of jamming it into the existing object. I'll do that.

danielkberry: Very nice, the new tests look great!

lib/metric-config-parser/metric_config_parser/experiment.py
danielkberry: Is the bucket config guaranteed to be provided by the API or are there cases in which it could be missing? I agree with the "if the bucket config is present, require the randomization unit to be valid"  rule

mikewilli: Honestly not sure. I'm guessing it will always be there (at least for experiments where we've made it this far and it matters for analysis), but I do tend to get tripped on these things by Normandy stuff sometimes. However, if the bucket_config isn't provided, we fall back to `None` which leads to the default behavior of `AnalysisUnit.CLIENT`. It's really the case where the bucket config exists, but the randomization_unit doesn't that I wasn't sure about. It seems pretty reasonable to consider that invalid, but maybe to confirm I can check in Experimenter next week to see if it's required, and base the decision here on that behavior (i.e., if it's not required by experimenter we could treat `None` as a valid option).

mikewilli: Ok confirmed with Jared -- randomization unit should never be None -- the randomization unit depends on the application, and an experiment must always have an application. So I think the current behavior (if bucket_config then randomization unit must not be None) is valid, and throwing an error if that expectation is not met should be fine.

danielkberry: Great, thanks for following up on this!

PR #579, opened 2024-09-04
create spotlight impression exposure event

florezita:


jetstream/1-click-set-to-default-existing-users.toml
```diff
@@ -0,0 +1,22 @@
...
+data_source = "spotlight_impressions"
+window_start = 0
+window_end = 7
+
+
+[data_sources]
+[data_sources.spotlight_impressions]
+from_expression = '''(SELECT *, metrics.uuid.messaging_system_client_id AS client_id,
+DATE(submission_timestamp) AS submission_date FROM `moz-fx-data-shared-prod.firefox_desktop.messaging_system` WHERE metrics.string.messaging_system_ping_type = 'spotlight')'''
+experiments_column_type = "native"
```
danielkberry, line 20: I think this should be `glean` not `native`. Although, I also think that the CI should've errored here

danielkberry (changes requested):

jetstream/1-click-set-to-default-existing-users.toml
florezita: maybe it didn't because of the `AS client_id` in `metrics.uuid.messaging_system_client_id AS client_id`?

florezita: i did run jetstream validate_config and it did throw an error when i didn't have that

danielkberry: These fields don't appear in the [dry run query](https://app.circleci.com/pipelines/github/mozilla/metric-hub/2327/workflows/23dc2c79-fe4b-4125-b8e3-206d63a0a3d5/jobs/4844) that we use to test the configuration

florezita: sounds good, i'll make the change

danielkberry:
Ok, talked with @mikewilli and we determined that the exposure signal ignores the data source's `experiments_column_type` parameter. So any values are valid here when used as the exposure signal but if this data source was used for a metric, then it should validate and error.

danielkberry (approved):
Thanks for the change!

PR #628, opened 2024-10-03
Update segment resolution in analysis configuration

RJAK11:
Modified ExperimentSpec.resolve to incorporate segments directly from the experiment object and include them in the analysis.


lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
```diff
@@ -267,6 +267,46 @@ def test_preenrollment(self, experiments, config_collection):
...
+            [segments.test_segment]
+            Data_Source = "test_data_source"
+            Select_Expression = "{{agg_any('1')}}"
+
+            [segments.data_sources.test_data_source]
+            from_expression = "(SELECT 1 WHERE submission_date BETWEEN {{experiment.start_date_str}} AND {{experiment.last_enrollment_date_str}})"
+            """  # noqa
+        )
+        spec = AnalysisSpec.from_dict(toml.loads(config_str))
+        cfg = spec.resolve(experiments[9], config_collection)
```
mikewilli, line 296: Great test! This actually raises an important question -- should the custom experiment config override or augment the default segments from Experimenter? This isn't an issue for outcomes because we don't support outcomes in custom experiment configs -- they're only defined via Experimenter. However, the precedent here that we can draw from is metric analysis periods, and in that case we would merge them from various sources to get a final list containing metrics from all sources (defaults, outcomes, custom). So for segments, we can do the same -- we'll always use the segments from Experimenter, and if there are segments in a custom config we'll add those on as well.

Mostly documenting this for posterity. We should probably also add this to the docs, though after a quick look I'm not sure exactly the best place for it. We can take a closer look later and see where makes the most sense.

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
```diff
@@ -267,6 +267,46 @@ def test_preenrollment(self, experiments, config_collection):
...
+
+        custom_segment = [seg for seg in cfg.experiment.segments if seg.name == "test_segment"]
```
mikewilli, line 298: Like above, I'd prefer that we assert the length of the segments, and then retrieve them individually and make further assertions. I'd also suggest the same regarding the data sources `from_expression` resolution.

I think asserting that `len(custom_segment) == 1` is still valid to ensure we're deduping properly, but it might be better to put that in a separate test where we add the `experiments[9]` segment to the list of custom experiment config segments.

mikewilli (changes requested):
Left a few suggested changes but overall great work! Thanks!

lib/metric-config-parser/metric_config_parser/experiment.py
```diff
@@ -290,11 +290,15 @@ def resolve(
         experiment: "Experiment",
         configs: "ConfigCollection",
     ) -> ExperimentConfiguration:
+
+        experiment_segments = [SegmentReference(seg) for seg in experiment.segments]
+        all_segments = self.segments + experiment_segments
```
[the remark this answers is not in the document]
RJAK11: Using set() gave the error `TypeError: unhashable type: 'SegmentReference'` so I instead checked for duplicates by iterating over the list of all segments

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
```diff
@@ -267,6 +267,88 @@ def test_preenrollment(self, experiments, config_collection):
...
+        cfg = spec.resolve(experiments[9], config_collection)
+
+        assert len(cfg.experiment.segments) == 1
+
+        segment = cfg.experiment.segments[0]
+
+        assert segment.name == "regular_users_v3"
+        assert segment.data_source.name == "my_cool_data_source"
+        expected_expression = "SELECT 1 WHERE submission_date BETWEEN"
+        assert expected_expression in segment.data_source.from_expression
```
mikewilli, line 282: I think you should be able to assert the full expression here to ensure that it resolves the templated parameters properly, something like (may have gotten the dates wrong):
```suggestion
        expected_expression = "SELECT 1 WHERE submission_date BETWEEN '2019-12-01' AND '2019-12-08"
        assert expected_expression == segment.data_source.from_expression
```

mikewilli: Also same comment for the other instances of this assertion.

lib/metric-config-parser/metric_config_parser/tests/test_experiment.py
mikewilli: Just needs an assertion on the total length of segments:
```suggestion
        assert len(cfg.experiment.segments) == 2

        custom_segment = [seg for seg in cfg.experiment.segments if seg.name == "test_segment"]
```

mikewilli (changes requested):
Almost there, just a few small test updates.

mikewilli (changes requested):
Looks good to me! Just remembered though -- need to bump the version to `2024.10.1`

mikewilli (approved):
:shipit: :tada:

PR #634, opened 2024-10-08
bug: cert_storage_memory should be histogram mean

mikewilli:
Experiment owner requested that we use a mean here because Jetstream doesn't support histogram metrics.

Tried to follow existing usage of `agg_histogram_mean`, but not totally sure if this is correct so please check my work here!

### DIFF
--- a/jetstream/outcomes/firefox_desktop/crlite.toml
+++ b/jetstream/outcomes/firefox_desktop/crlite.toml
@@ -3,10 +3,10 @@ description = "Metrics for the CRLite certificate revocation checking system"
 
 [metrics.cert_storage_memory]
 data_source = "metrics"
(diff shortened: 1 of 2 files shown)


jetstream/outcomes/firefox_desktop/crlite.toml
```diff
@@ -3,10 +3,10 @@ description = "Metrics for the CRLite certificate revocation checking system"
 
 [metrics.cert_storage_memory]
 data_source = "metrics"
-select_expression = "ARRAY_AGG(metrics.memory_distribution.cert_storage_memory IGNORE NULLS)"
+select_expression = "{{agg_histogram_mean('metrics.memory_distribution.cert_storage_memory)'}}"
```
scholtzan, line 6:
```suggestion
select_expression = "{{agg_histogram_mean('metrics.memory_distribution.cert_storage_memory')}}"
```

PR #637, opened 2024-10-09
Add ad_marketplace metrics to ads.toml

ryboland:
Create metrics from `definitions/` to `looker/definitions` file in order to surface these metrics in Looker


mikewilli:
I'm going to close this since it's 1.5yrs old, but feel free to re-open and update if needed.

PR #640, opened 2024-10-10
bug: pop ratio metrics must specify depends_on

mikewilli:
Getting an error for both of these that `revocation_check_counter` doesn't exist, so guessing we just need to make sure that the `depends_on` is set so the numerator/denominator are available before computing this metric's statistic.

### DIFF
--- a/jetstream/outcomes/firefox_desktop/crlite.toml
+++ b/jetstream/outcomes/firefox_desktop/crlite.toml
@@ -32,6 +32,7 @@ select_expression = """
 """
 friendly_name = "CRLite usage"
 description = "Number of revocation checks performed using CRLite."
+depends_on = ["crlite_use_counter", "revocation_check_counter"]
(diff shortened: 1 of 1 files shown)


jetstream/outcomes/firefox_desktop/crlite.toml
```diff
@@ -48,11 +53,18 @@ select_expression = """
 """
 friendly_name = "OCSP usage"
 description = "Number of revocation checks performed using OCSP."
+depends_on = ["ocsp_use_counter", "revocation_check_counter"]
```
scholtzan, line 56:
```suggestion
```

this can go I guess with the `_ratio` metric

mikewilli: Ah whoops, thanks!

scholtzan (approved):

PR #658, opened 2024-10-23
Implement segment retrieval function in metric-config-parser

RJAK11:
Added a new function in metric-config-parser that retrieves and returns all available segments based on application.

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/config.py
+++ b/lib/metric-config-parser/metric_config_parser/config.py
@@ -680,6 +680,15 @@ class ConfigCollection:
 
         return None
 
+    def get_all_app_segments(self, app_name: str) -> Optional[List[SegmentDefinition]]:
+        segments: List[SegmentDefinition] = []
+        for definition in self.definitions:
+            if app_name == definition.platform:
+                if not isinstance(definition.spec, MonitoringSpec):
+                    segments.extend(definition.spec.segments.definitions.values())
+
+        return segments
+
     def get_metrics_sql(
         self,
         metrics: List[str],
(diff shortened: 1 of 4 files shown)


lib/metric-config-parser/metric_config_parser/tests/test_config.py
```diff
@@ -850,3 +850,40 @@ def test_merge_with_wildcards_invalid(self):
...
+            platform="firefox_desktop",
+            spec=AnalysisSpec.from_dict(toml.loads(config_str)),
+            last_modified=datetime.datetime.now(),
+        )
+
+        config_collection = ConfigCollection(
+            configs=[],
+            outcomes=[],
+            defaults=[],
+            definitions=[definition],
```
mikewilli, line 881: Can we add another DefinitionConfig to this list with some non-desktop segments? Then we'd also be validating the filtering works properly at the same time.

mikewilli (approved):
r+wc Looks good, just a few suggestions. :tada:

lib/metric-config-parser/metric_config_parser/config.py
```diff
@@ -680,6 +680,15 @@ def get_segment_definition(self, slug: str, app_name: str) -> Optional[SegmentDe
 
         return None
 
+    def get_all_app_segments(self, app_name: str) -> Optional[List[SegmentDefinition]]:
+        segments: List[SegmentDefinition] = []
+        for definition in self.definitions:
+            if app_name == definition.platform:
+                if not isinstance(definition.spec, MonitoringSpec):
```
mikewilli, line 687: Oh also, these could just be combined into a single check.

PR #683, opened 2024-11-15
feat: create ConfigCollection from non-git path

mikewilli:
* Adds a `from_local_path` function to `ConfigCollection`
* bumps version to 2024.11.1

Why? Experimenter wants to use metric-config-parser to parse a partial local copy of metric-hub for outcomes and segments, and the existing `from_local_repo` required some mock objects in order for the local copy to work as a git repo (Experimenter always wants the latest version, so has no need for the `.git` dir that is used to get configs from a specific date, and `.git` is kind of big!).

Related: https://github.com/mozilla/experimenter/pull/11772

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/config.py
+++ b/lib/metric-config-parser/metric_config_parser/config.py
@@ -494,6 +494,101 @@ class ConfigCollection:
             is_private=is_private,
         )
 
+    @classmethod
+    def from_local_path(cls, path: str, is_private: bool = False) -> "ConfigCollection":
+        """Load configs from a local non-repo copy of a metric-hub-like folder structure.
+
+        Args:
+            path (str): path to the configs. Looks for TOML files in the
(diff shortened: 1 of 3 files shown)


RJAK11 (approved):
That looks good to me!

scholtzan (approved):
I guess this should be fine. One thing to consider though is that some class methods, like `as_of()` won't work as they rely on a repository being present. Maybe we need to add some better error handling for those, but I guess experimenter won't look at the commit history

mikewilli:
@scholtzan Thanks for raising this valid concern. I updated the PR to create a new sub-class of `ConfigCollection` that explicitly raises an exception if the class methods that use repos are called, and then moved the `from_local_path` function into this class.

Of note, I think it sort of makes more sense the other way around, as ConfigCollection adds repos to the LocalConfigCollection, but I implemented it this way to avoid making changes in ConfigCollection (just in case).

PR #690, opened 2024-11-19
[ci skip-rerun] fix: change desktop default DAU metric to v2

mikewilli:
* change the default DAU metric for experiments to the non-deprecated `_v2` version
* also matches with mobile, which is helpful for changing the guardrails in Experimenter
* skip rerunning jetstream for all desktop experiments


bochocki (approved):
lgtm!

PR #714, opened 2024-12-16
Example config for monitoring Glean network http errors

scholtzan:
See https://mozilla.slack.com/archives/C4D5ZA91B/p1733998881469899

### DIFF
--- a/opmon/firefox-desktop-networking.toml
+++ b/opmon/firefox-desktop-networking.toml
@@ -1,0 +1,73 @@
+[project]
+
+name = "Firefox Desktop Networking Monitoring"
+platform = "firefox_desktop"
+xaxis = "submission_date"
+start_date = "2024-12-05"
+# We want continuous monitoring of this data
+#end_date =
+skip_default_metrics = true
+
+metrics = [
+  "total_network_close_errors"
+]
+
+alerts = [
+  "historical_diff_total_network_close_errors"
+]
+
+
+[project.population]
+
+data_source = "network_close_reasons"
+monitor_entire_population = true
+dimensions = ["network_close_reason"]
+group_by_dimension = "network_close_reason"
+
+
+[metrics]
+
+[metrics.total_network_close_errors]
+data_source = "network_close_reasons"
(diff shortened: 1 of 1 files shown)


opmon/firefox-desktop-networking.toml
```diff
@@ -0,0 +1,73 @@
...
+  "historical_diff_total_network_close_errors"
+]
+
+
+[project.population]
+
+data_source = "network_close_reasons"
+monitor_entire_population = true
+dimensions = ["network_close_reason"]
+group_by_dimension = "network_close_reason"
```
scholtzan, line 25: This will create separate graphs and alerts for every error type/key value in `networking_http_3_connection_close_reason`

opmon/firefox-desktop-networking.toml
```diff
@@ -0,0 +1,73 @@
...
+[metrics]
+
+[metrics.total_network_close_errors]
+data_source = "network_close_reasons"
+select_expression = "SUM(total_count)"
+friendly_name = "Total network close errors"
+description = "Total number of network close errors"
+type = "scalar"
+
+[metrics.total_network_close_errors.statistics.sum]
```
scholtzan, line 37: This will sum up all the error counts across all clients for each error type

opmon/firefox-desktop-networking.toml
```diff
@@ -0,0 +1,73 @@
...
+
+[alerts]
+
+[alerts.historical_diff_total_network_close_errors]
+type = "avg_diff"
+metrics = [
+  "total_network_close_errors"
+]
+window_size = 7
+max_relative_change = 0.5
```
scholtzan, line 73: This alert will compare the most recent 7 days of values to the 7 days prior to that. If the relative change is more than 50%, then it will trigger an alert. The settings here might need to be tuned. Alerts might also get triggered when user counts are down, e.g. during holidays.

scholtzan:
Implemented in Redash: https://mozilla.slack.com/archives/C4D5ZA91B/p1733998881469899

PR #720, opened 2024-12-20
[ci rerun-skip] add default-ui-experiment config to end analysis now

mikewilli:
Automated analysis results are not being used for this experiment, so let's just artificially end the experiment so we don't need to run analysis anymore.

- ci rerun-skip so we don't do a rerun for the stuff we've already computed


danielkberry (approved):
LGTM, thanks!

PR #730, opened 2025-01-09
Create relay.toml

pissac17:
outcome metrics for relay integration
JIRA: https://mozilla-hub.atlassian.net/browse/DS-3948
Experiment brief: https://docs.google.com/document/d/1lfVZUKUHQKUuj0IYt-2Uzrui8TYRBJ8ZBn6AKBgIF7A/edit?tab=t.0

### DIFF
--- a/jetstream/outcomes/firefox_desktop/relay.toml
+++ b/jetstream/outcomes/firefox_desktop/relay.toml
@@ -8,8 +8,8 @@ friendly_name = "Relay offer shown to client"
 description = "Number of times Relay offer is shown"
 select_expression = """
       COALESCE(COUNTIF(
-         events.name= 'shown_offer_relay'
-         AND events.category = 'relay_integration'
+         event_method= 'shown_offer_relay'
+         AND event_category = 'relay_integration'
       ),0)
 """
 data_source = "events"   
@@ -20,8 +20,8 @@ friendly_name = "Relay offer clicked by client"
 description = "Number of times user clicked on Relay offer"
 select_expression = """
       COALESCE(COUNTIF(
-         events.name= 'clicked_offer_relay'
-         AND events.category = 'relay_integration'
+         event_method = 'clicked_offer_relay'
+         AND event_category = 'relay_integration'
       ),0)
 """
 data_source = "events"   
@@ -32,8 +32,8 @@ friendly_name = "Relay optin panel shown to client"
 description = "Number of times Relay optin panel is shown"
 select_expression = """
       COALESCE(COUNTIF(
-         events.name= 'shown_opt_in_panel'
-         AND events.category = 'relay_integration'
+         event_method = 'shown_opt_in_panel'
+         AND event_category = 'relay_integration'
       ),0)
 """
 data_source = "events"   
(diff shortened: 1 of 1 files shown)


jetstream/outcomes/firefox_desktop/relay.toml
```diff
@@ -0,0 +1,120 @@
...
+
+# Metric for tracking how many times the offer was shown
+[metrics.relay_integration_shown_relay_offer]
+friendly_name = "Relay offer shown to client"
+description = "Number of times Relay offer is shown"
+select_expression = """
+      COALESCE(COUNTIF(
+         event_method= 'shown_offer_relay'
+         AND event_category = 'relay_integration'
+      ),0)
```
groovecoder, line 13: change/suggestions (blocking): If these outcome queries pull from the same `events` source as the sql.telemetry.mozilla.org queries' `telemetry.events`, then this query needs to change to:
```suggestion
      COALESCE(COUNTIF(
         event_method = 'shown'
         AND event_object = 'offer_relay'
         AND event_category = 'relay_integration'
      ),0)
```

You can see the existing sql.telemetry.mozilla.org relay integration opt-in conversion funnel query here: https://sql.telemetry.mozilla.org/queries/93179/source#230436

jetstream/outcomes/firefox_desktop/relay.toml
```diff
@@ -0,0 +1,120 @@
...
+"""
+data_source = "events"   
+statistics = { bootstrap_mean = {}, deciles = {} }
+
+[metrics.relay_integration_shown_fill_username]
+friendly_name = "Relay autocomplete popup"
+description = "Fraction of Users that return to original form and receive an email mask"
+select_expression = """
+       COALESCE(MAX(
```
groovecoder, line 82: question (non-blocking): why is this `MAX` when all the others are `COUNTIF` ?

groovecoder: change & question (blocking): With a fresh new year's look into the code, I realize that the `fill_username` object is not what I thought it was. 😢 I thought the code fired this event when the user came back to the original form from which they started the Relay opt-in flow, but it doesn't. This event fires for users who have already opted into Relay and then use Relay again.

So this one isn't part of the "opt into Relay" conversion funnel - it's the "keep using Relay" engagement metric. Is there a way to group these outcomes into 3 groups:

1. Opt into Relay conversion funnel
2. Relay engagement
3. Upgrade to Premium Relay conversion funnel

?

groovecoder (changes requested):
As mentioned in the individual PR comment, if these outcome queries pull from the same `events` source as the sql.telemetry.mozilla.org queries' `telemetry.events`, then we need to change all the select expressions to use the `event_object` field as suggested for the `relay_integration_shown_relay_offer` metric.

jetstream/outcomes/firefox_desktop/relay.toml
pissac17: Understood. Would you like me to remove the 'fill username' metric from this outcome file? As for your question about dividing the outcomes into three groups, I’m not sure if that’s possible. This might be something @shell1 can help with. @shell1, could you provide some guidance on this?

pissac17: @groovecoder Regarding the use of MAX instead of COUNTIF for 'fill username' : MAX is suitable when you need a binary indicator (e.g., "Did the condition occur at least once?"). Based on my understanding of the experiment brief, the focus is on calculating the _fraction of users_ who return to the original form and receive an email mask. However, if the goal is to measure the total number events, COUNTIF would be the appropriate choice. Could you clarify what you are trying to measure from this metric?

groovecoder: Yes, we're most interested in the fraction of users for the experiment outcome. If we need to look at total number of events (e.g., for a question like "How many times did a client see the Relay offer before clicking it?), we'll use another query.

PR #734, opened 2025-01-12
Improving native ads reporting v2

sergiosonline:
* Fixing naming of ratio metrics
* Masking metrics from appearing as dimensions

### DIFF
--- a/looker/definitions/ads.toml
+++ b/looker/definitions/ads.toml
@@ -30,48 +30,45 @@ description = "Daily SPOC revenue"
 
 [metrics.native_spend]
 select_expression = "SUM(spend)"
-data_source = "native_desktop_ad_metrics"
+data_source = "native_desktop_ad_metrics_hourly"
 friendly_name = "Native Spend"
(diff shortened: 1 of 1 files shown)


looker/definitions/ads.toml
```diff
@@ -139,22 +149,32 @@ from_expression = """
...
+    campaign_id
   FROM `moz-fx-data-shared-prod.ads.native_desktop_ad_metrics_hourly`
 )
 """
 friendly_name = "Native Desktop Ads Metrics Daily"
 description = """
   Tracks Native Desktop Ad Metrics Across Days
 """
-columns_as_dimensions = true
+columns_as_dimensions = false
```
scholtzan, line 160:
```suggestion
columns_as_dimensions = true
```

I believe this should be `true`? The columns of this data source should show up in Looker, right?

sergiosonline: Correct. Thanks for flagging.

scholtzan (approved):

PR #735, opened 2025-01-13
[ci rerun-skip] certificate-transparency: add exp filter to tls_metrics ds

mikewilli:
Just tested this experiment slug filter and it seemed to run ok, so let's try this.

Also, doing rerun-skip and I'll just manually rerun the couple of dates that failed last time.


betling (approved):
LGTM

PR #737, opened 2025-01-14
[ci rerun-skip] certificate-transparency: use legacy id for tls timing

betling:
changed tls timing data source to use legacy client id in experiment toml file.  also removed division in metric 'from' statement since that is what worked in our test query.

### DIFF
--- a/jetstream/certificate-transparency.toml
+++ b/jetstream/certificate-transparency.toml
@@ -17,8 +17,8 @@ data_source = "events_certerror"
 
 [metrics.tls_timing]
 friendly_name = "TLS timing"
-description = "Amount of time for TLS handshake in miliseconds"
-select_expression = "COALESCE(SUM(CAST(key AS INT64)), 0)"
+description = "Amount of time for TLS handshake in seconds"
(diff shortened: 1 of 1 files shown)


jetstream/certificate-transparency.toml
```diff
@@ -17,8 +17,8 @@ data_source = "events_certerror"
 
 [metrics.tls_timing]
 friendly_name = "TLS timing"
-description = "Amount of time for TLS handshake in seconds"
-select_expression = "COALESCE(SUM(CAST(key AS INT64)/1000000), 0)"
+description = "Amount of time for TLS handshake in miliseconds"
+select_expression = "COALESCE(SUM(CAST(key AS INT64)), 0)"
```
mikewilli, line 21: I think you should be safe to leave this alone, I was just testing some different things and forgot to change it back. If you do decide to leave it as-is, pretty sure it should be microseconds.

```suggestion
description = "Amount of time for TLS handshake in seconds"
select_expression = "COALESCE(SUM(CAST(key AS INT64)/1000000), 0)"
```

mikewilli (approved):
I think you are fine to leave the microseconds to seconds conversion, but LGTM thanks!

PR #740, opened 2025-01-16
[ci rerun-skip] certificate-transparency: add timing ratio metrics and deciles

betling:
add number of handshakes and timing ratio metrics, aggregate statistics daily, and also add deciles.

### DIFF
--- a/jetstream/certificate-transparency.toml
+++ b/jetstream/certificate-transparency.toml
@@ -1,7 +1,8 @@
 [metrics]
 
-weekly = ["cert_errors", "tls_timing"]
-overall = ["cert_errors", "tls_timing"]
+daily = ["cert_errors", "tls_timing", "num_handshakes", "timing_ratio"]
+weekly = ["cert_errors", "tls_timing", "num_handshakes", "timing_ratio"]
+overall = ["cert_errors", "tls_timing", "num_handshakes", "timing_ratio"]
 
 [metrics.cert_errors]
 friendly_name = "TLS certification errors"
@@ -14,6 +15,7 @@ select_expression = """(
 data_source = "events_certerror"
 
 [metrics.cert_errors.statistics.bootstrap_mean]
+[metrics.cert_errors.statistics.deciles]
 
 [metrics.tls_timing]
 friendly_name = "TLS timing"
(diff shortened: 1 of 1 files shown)


jetstream/certificate-transparency.toml
```diff
@@ -22,6 +24,25 @@ select_expression = "COALESCE(SUM(CAST(key AS INT64)/1000000), 0)"
...
+description = "Number of handshakes computed by counting number of values in metrics.timing_distribution.network_tls_handshake.values"
+select_expression = "COALESCE(COUNT(key), 0)"
+data_source = "tls_metrics"
+
+[metrics.num_handshakes.statistics.bootstrap_mean]
+[metrics.num_handshakes.statistics.deciles]
+
+[metrics.timing_ratio]
+friendly_name = "TLS timing ratio"
+description = "TLS timing divided by number of handshakes"
```
mikewilli, line 40:
```suggestion
description = "TLS timing (in seconds) divided by number of handshakes"
```

jetstream/certificate-transparency.toml
```diff
@@ -1,7 +1,8 @@
 [metrics]
 
-weekly = ["cert_errors", "tls_timing"]
-overall = ["cert_errors", "tls_timing"]
+daily = ["cert_errors", "tls_timing", "num_handshakes", "timing_ratio"]
```
mikewilli, line 3: Do you need `cert_errors` in the daily analysis as well? This will add another additional data source to every daily metric query, which will be slower and cost more. Not necessarily a problem, but just FYI.

mikewilli (approved):
So this looks correct to me so I'm approving, but a few caveats/warnings that I think are worth mentioning:

- the deciles may not be exactly what you're expecting: unlike GLAM, Jetstream's quantiles are over users, not over the event itself
  - most of the time when people want quantiles, they are expecting quantiles in the metric itself like GLAM
  - additionally, we don't display deciles stats in Experimenter
- this will likely take a while to compute (each daily metric query now has 2 additional data sources and 4 additional metrics to compute, plus each analysis period has new metrics and statistics)
  - we may have to iterate on it to reduce resource consumption if there are issues with analysis completing


Also, I left a comment about this, but consider whether all of these custom metrics need to be computed daily or if some of them can be left for weekly/overall only.

jetstream/certificate-transparency.toml
betling: i've removed the daily and weekly aggregations in latest commit so it is just overall.

mikewilli (approved):
LGTM! If you're just doing `overall`, go ahead and add `[ci rerun-skip]` to the PR title and I'll just run it manually for you when it's merged.

PR #750, opened 2025-02-06
add metric to pdf outcomes and rerun two experiments

irrationalagent:
@danielkberry - the dates in the experiment config are the same as what's in experimenter - will that still trigger a re-run?


danielkberry (approved):

danielkberry:
Yeah, Jetstream isn't smart enough to decide what's a "meaningful" change or not, any change to a config triggers a rerun

PR #774, opened 2025-02-20
feat: add special_onboarding_events_v2 ds using events_stream

mikewilli:
promote-add-ons-in-onboarding-release had an issue with resources during analysis, and I noticed this datasource includes two UNNESTs and a JOIN within the expression. I created a new DS using events_stream instead of events, which removes the need for the UNNESTs, and drastically reduces data scanned.

Not sure if this will completely resolve the issue, but it should at least make for a much simpler query for the affected metrics, and we can iterate from there.

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -443,28 +443,28 @@ owner = "loines@mozilla.com"
 
 [metrics.turn_on_notifications_ctr_onboarding]
 select_expression = "COALESCE(SUM(turn_on_notifications_flag))"
-data_source = "special_onboarding_events"
+data_source = "special_onboarding_events_v2"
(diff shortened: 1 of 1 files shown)


danielkberry (approved):
Definitely like the idea of pushing folks to adopt the `events_stream` tables. Any opinion on naming the data source in a way that reflects that it's just pointing at the stream table instead of the original one? `_v2` might imply that the previous version is deprecated or out of date.

mikewilli:
I was thinking about the naming for exactly the reasons you mention, and was originally going to name the new one by just appending `_stream` to the name since it already ends with `events`. However, I decided that this should probably supercede usage of the current data source since it's the same data and I don't think there are any downsides (unless I'm wrong about one of those?).

PR #779, opened 2025-02-28
chore: docs on CI reruns

mikewilli:
I wrote up some info on how `[ci rerun-skip]` works and why/when you may or may not want to use it. Any feedback requested, especially to make sure it's clear, (relatively) comprehensive, and not overly complicated.


README.md
```diff
@@ -13,3 +13,51 @@ All metric definitions can and should be referenced in other tooling and their c
...
+metric-hub's CI is configured to detect when a given change would affect experiment analysis, and to automatically rerun analysis for these experiments in certain scenarios. This behavior is not always desirable, and so it can also be overridden using `[ci rerun-skip]`. Continue reading to learn more about the automated rerun scenarios and when you may want to use `[ci rerun-skip]`.
+
+### Automated Analysis Reruns
+
+There are two main scenarios that trigger analysis reruns:
+
+1. Changing default metrics (i.e., in `jetstream/defaults`)
+    * this will trigger a rerun in the affected application of:
+      * **all live** experiments
+      * **completed** experiments that have **ended within the past 90 days**
```
danielkberry, line 29: TIL, I thought it was only live experiments.

danielkberry (approved):
LGTM, thanks for writing this up!

README.md
scholtzan: Should we change it to live experiments only?

scholtzan (approved):

README.md
danielkberry: IMO, yeah, we should.  I think we tend to follow a pattern of "avoid reanalyzing and thus potentially changing past experiment results unless explicitly requested" and this seems to break that.

mikewilli: Yea unless I'm mistaken (and it is late on a Friday, so...) the check for that is here: https://github.com/mozilla/jetstream/blob/main/jetstream/cli.py#L414-L415.

Since this function is used by more than just the CI rerun functionality, we should make sure this won't affect anything else before changing, but in general it should be pretty straightforward to do.

README.md
shell1, line 1: Skip:  
  * if you only care about changes going forward
    * if the only thing changing is statistics
    * if you only care about one analysis window (e.g., overall)
    
    I don't understand lines 38+39.  what does "changes going forward" or statistics mean?
    skip to avoid the cost/time of running the extra analysis for weekly details, if you only need the overall results
    skip if you only need new metrics added that don't have weekly calculations.

shell1: nevermind - i read further and see that "going forward" would just mean it is added for the overall but won't rerun the already analyzed days with the changed file.

PR #784, opened 2025-03-04
Custom config: Higher placement phase 3 relaunch

sergiosonline:
Adding based on https://github.com/mozilla/metric-hub/blob/main/jetstream/sponsored-suggestion-placement-phase-2.toml

A few things to note:
* This phase 3 was done with fixed enrolment and so no need to examine cohorts.
* Exposure events


jetstream/higher-placement-phase-3-relaunch.toml
```diff
@@ -0,0 +1,622 @@
...
+select_expression = "COUNT(DISTINCT IF(ad_component AND num_ads_loaded_reported > 0, impression_id, NULL))"
+data_source = "tagged_google_serp_events"
+friendly_name = "SERPs with ads"
+description = "Tagged Google SERP page loads with ads loaded"
+
+[metrics.serp_impressions_with_ads.statistics.bootstrap_mean]
+[metrics.serp_impressions_with_ads.statistics.deciles]
+
+[metrics.serp_has_ad_click]
+select_expression = "COUNT(DISTINCT IF(ad_component AND num_ads_showing > 0 AND num_clicks > 0, impression_id, NULL))"
```
pissac17, line 143: the field `num_ads_showing` is no longer in use and should be replaced with `num_ads_visible`. You can find the updated field names here: https://docs.telemetry.mozilla.org/datasets/other/serp_events/reference

jetstream/higher-placement-phase-3-relaunch.toml
```diff
@@ -0,0 +1,622 @@
...
+      SELECT
+        glean_client_id,
+        ping_seq AS seq,
+        event_timestamp,
+        sap_source,
+        legacy_telemetry_client_id,
+        impression_id,
+        submission_date,
+        ANY_VALUE(experiments) AS experiments,
+        LOGICAL_OR(ad_component AND num_ads_loaded_reported > 0) AS has_ads_loaded,
```
pissac17, line 518: It looks like this query is using outdated field names and should be updated to reflect the current ones listed here: [SERP Telemetry Documentation](https://docs.telemetry.mozilla.org/datasets/other/serp_events/reference). For example, `num_ads_loaded_reported` should be replaced with `num_ads_loaded`

jetstream/higher-placement-phase-3-relaunch.toml
```diff
@@ -0,0 +1,622 @@
...
+        LOGICAL_OR(ad_component AND num_ads_showing > 0 AND num_clicks > 0) AS has_ad_click,
+        SUM(IF(ad_component AND num_ads_showing > 0, num_clicks, 0)) AS num_ad_clicks,
+      FROM (
+        -- redefine ad componet to exclude non-monetized components
+        SELECT
+          * EXCEPT (is_ad_component),
+          component IN ('ad_carousel', 'ad_image_row', 'ad_link', 'ad_sidebar', 'ad_sitelink') AS ad_component
+        FROM
+          `moz-fx-data-shared-prod.firefox_desktop_derived.serp_events_v1`
```
pissac17, line 527: I think `moz-fx-data-shared-prod.firefox_desktop_derived.serp_events_v1` table is no longer in use, the updated data can be found in `moz-fx-data-shared-prod.firefox_desktop_derived.serp_events_v2` .

sergiosonline: Thanks, @pissac17 . Does it make sense to use the view `serp_events` here since it leverages UDFs to compute a lot of the fields we're interested in measuring, e.g., `num_ads_loaded`, `num_clicks`? Have you worked on any recent analyses that show how to map the old v1 table to the new v2?

pissac17: @sergiosonline Yes, I would recommend using the `serp_events` table here instead of the raw table. The underlying table for `serp_events`  is `moz-fx-data-shared-prod.firefox_desktop_derived.serp_events_v2`, but serp_events is just easier to use.
I haven't done any analysis on mapping between the v1 and v2 tables. You could use this config file for reference though: [yelp_suggestion_config](https://github.com/mozilla/metric-hub/blob/57cd56a2fee4ed441a172a7c6cfac10a45d3fb3e/jetstream/yelp-suggestions-pilot-v3.toml#L4)

sergiosonline: @m-d-bowerman , @pissac17 : I took a stab at remapping the data source query to match the latest serp_events table [here](https://console.cloud.google.com/bigquery?sq=470293616514:f199cd07e4a94890bcd98a828b69b283).

Couple of things:
- A major issue I see in the output is that it shows clients with either missing exposure info from urlbar_fields and complete serp_fields interaction data, or missing exposure info from urlbar_fields and missing serp_fields interaction data. Am I missing something in the way these two data sources are combined?
- I see multiple paradigms in drafting these custom config files: Is there a preference between stating a custom query as a data source explicitily in this file, vs creating a personal derived table and pulling the info directly from it?
- Do segments need to be defined if we go ahead in defining the exposure_signal in the custom config? Or does that exposure signal get applied to all metrics automatically?

m-d-bowerman: >     * A major issue I see in the output is that it shows clients with either missing exposure info from urlbar_fields and complete serp_fields interaction data, or missing exposure info from urlbar_fields and missing serp_fields interaction data. Am I missing something in the way these two data sources are combined?
Clients will only show up in `serp_events` if they issued a search and landed on a SERP, which would explain how they could be in `urlbar_events` but not `serp_events`. Being in `serp_events` and not in `urlbar_events` is a bit weirder, but I could see it happening because of details around ping scheduling or if clients are visiting search engine sites organically and not through the URL bar.
>     * I see multiple paradigms in drafting these custom config files: Is there a preference between stating a custom query as a data source explicitily in this file, vs creating a personal derived table and pulling the info directly from it?
Dave did it in the past to cut down on the query size for Jetstream, filtering a data source first rather than querying the entire table at analysis time.
>     * Do segments need to be defined if we go ahead in defining the exposure_signal in the custom config? Or does that exposure signal get applied to all metrics automatically?
No, we don't need segments here -- the exposure signal will apply to all metrics when you change the analysis basis to exposure rather than enrollment.

jetstream/higher-placement-phase-3-relaunch.toml
```diff
@@ -0,0 +1,639 @@
...
+      submission_date,
+      -- experiments,
+      FALSE AS is_serp,
+      STRUCT(
+        exposed,
+        event_action,
+        product_selected_result,
+        (
+          SELECT
+            COUNTIF(product_result_type = 'admarketplace_sponsored') > 0
```
m-d-bowerman, line 572: This is the wrong `product_result_type` for the exposure. They tracked `rust_adm_sponsored`, `rs_adm_sponsored`, and `merino_adm_sponsored`

jetstream/higher-placement-phase-3-relaunch.toml
```diff
@@ -0,0 +1,639 @@
...
+      is_terminal
+      AND submission_date BETWEEN '2025-01-27' AND '2025-02-18'
+      AND in_experiment(experiments, sample_id)
+  ),
+  urlbar_with_exposed AS (
+    SELECT
+      glean_client_id,
+      seq,
+      event_timestamp,
+      (select logical_or(r.product_result_type = 'admarketplace_sponsored') from unnest(results) as r) as exposed,
```
m-d-bowerman, line 519: See below regarding the exposure signal

jetstream/higher-placement-phase-3-relaunch.toml
sergiosonline: @m-d-bowerman : The above exposure signals as seen in the Experiment config aren't flowing down as a product_result_type, but `admarketplace_sponsored` is. [Query](https://console.cloud.google.com/bigquery?sq=470293616514:14597eaa5c48440389888d65bae4a68e).

I'm not super familiar with `mozdata.firefox_desktop.urlbar_events` and some of the other data sources, but I believe that we have a udf `bigquery-etl/sql/mozfun/norm/result_type_to_product_name/udf.sql` that maps values `rust_adm_sponsored`, `rs_adm_sponsored`, and `merino_adm_sponsored` ->  `admarketplace_sponsored`. Here's a doc detailing my thoughts: https://docs.google.com/document/d/1XrAsCy-dGjdN38NKuUesk3xNVYIbjfuE3sseD7cHojA/edit?usp=sharing

Appreciate some confirmation as to whether this is right.

m-d-bowerman:
You need to persist the `experiments` field in those data sources for Jetstream (it'll always look for that field to determine if a client's enrolled).

m-d-bowerman (approved):
Looks good. I'm not familiar with these experiments, though, so someone else should check if the included metrics cover everything asked for by the experiment owner.

pissac17 (approved):
LGTM

PR #813, opened 2025-03-26
Update device-migration-accounts-toolbar-icon.toml

pissac17:
adding segments for users that were signed into FxA pre enrollment, vs the ones that signed in post enrollment


florezita (approved):
lgtm

PR #830, opened 2025-03-31
add custom config for success metrics

florezita:
https://mozilla-hub.atlassian.net/browse/DS-4216

https://docs.google.com/document/d/1LuF0zYrnOQQThrAsJUcj7_SirJx-zu0tyuv0vIAgUfc/edit?tab=t.0

I added two metrics to this custom config. Some questions:

1. `search_engine_default_change` metric uses glean events stream as a data source, should I filter this data source further to `event_category = 'search.engine.default'` for computational purposes?

2. for the metric `non_default_search_engine_search` I had to UNNEST a field in the metrics glean table and denote it as a data source, is this viable?

Thanks!

### DIFF
--- a/jetstream/address-bar-update-donoharm.toml
+++ b/jetstream/address-bar-update-donoharm.toml
@@ -1,0 +1,42 @@
+[experiment]
+end_date = "2025-03-31"
+
+[metrics]
+weekly = ['search_engine_default_change', 'non_default_search_engine_search']
+overall = ['search_engine_default_change', 'non_default_search_engine_search']
+
+
+[metrics.search_engine_default_change]
+friendly_name = "Search Engine Default Changed"
+description = "If the Search Engine Default Changed"
+select_expression = """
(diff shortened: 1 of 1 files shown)


mikewilli:
I know you've already merged it, but I'll answer your questions anyway! The metrics query timed out in BigQuery so will need to optimize a bit.

>search_engine_default_change metric uses glean events stream as a data source, should I filter this data source further to event_category = 'search.engine.default' for computational purposes?

Yea I'd always recommend filtering `events_stream` by `event_category` as this is a clustering field and should help make the metrics query more efficient. And definitely worth going back to do it in this case since the metrics query timed out.

>for the metric non_default_search_engine_search I had to UNNEST a field in the metrics glean table and denote it as a data source, is this viable?

It should work, but I assume you mean viable from a performance perspective? If so then I'm guessing it'd just depends on the scale of data you're unnesting and how many clients are enrolling. Also consider using `CROSS JOIN` syntax so it's more explicit what you're trying to do (but this is more of a personal preference -- they are equivalent).

PR #836, opened 2025-04-03
Add Webcompat JetStream Outcome

bvandersloot-mozilla:
This is similar to Strict PBM, but has broken site reporter tool telemetry.

### DIFF
--- a/jetstream/outcomes/firefox_desktop/webcompat.toml
+++ b/jetstream/outcomes/firefox_desktop/webcompat.toml
@@ -1,0 +1,70 @@
+friendly_name = "Webcompat"
+description = """
+Metrics typically influenced by webcompat-breakages.
+Currently contains ETP Disablement, URI reloads,
+webcompat reporting panel impressions,
+and webcompat reporting panel uses.
+"""
+
+[metrics.etp_disablement]
+select_expression = "COALESCE(COUNTIF(event_name  = 'click_etp_toggle_off'), 0)"
+data_source = "protections_popup_events"
+friendly_name = "ETP Disablement"
+
+[metrics.etp_disablement.statistics.bootstrap_mean]
+
+
+[metrics.uri_reloads]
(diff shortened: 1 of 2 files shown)


jetstream/outcomes/firefox_desktop/webcompat.toml
```diff
@@ -0,0 +1,70 @@
...
+[metrics.etp_disablement]
+select_expression = "COALESCE(COUNTIF(event_name  = 'click_etp_toggle_off'), 0)"
+data_source = "protections_popup_events"
+friendly_name = "ETP Disablement"
+
+[metrics.etp_disablement.statistics.bootstrap_mean]
+
+
+[metrics.uri_reloads]
+select_expression = 'COALESCE(SUM(payload.processes.parent.scalars.browser_engagement_tab_reload_count), 0)'
```
bvandersloot-mozilla, line 18: This also has a Glean ping... I should probably convert that too.

scholtzan:
This looks right to me. Creating a separate datasource to unnest and filter events is a pretty common pattern (e.g. also done here: https://github.com/mozilla/metric-hub/blob/1c44999acff53d981962f0f10bad3635ffba19e6/jetstream/navigational-suggestions.toml#L128-L143)

An alternative that would not require manual unnesting would be to use the `events_stream` view instead. But a custom datasource would still be needed to filter on the `event_category` in order save cost and query runtime (since the table is clustered on category)

bvandersloot-mozilla:
I want to use this in OpMon as well- would it make sense to just put the metric definitions into `definitions/firefox_desktop.toml`? Also, I think at least some of these would be candidates for `opmon/definitions/firefox_desktop.toml`, but I can make that in a separate PR.

I'll convert it to use event_stream to not do the manual unnest- I didn't notice that events_stream is actually what I wanted!

scholtzan (approved):

PR #837, opened 2025-04-04
Adding custom campaign name

sergiosonline:
Using regexp to extract custom_campaign_name. Not sure if Looker will take kindly to a data source update given all the dashboards and dependencies we've baked in...


jaredsnyder (approved):
Just adds the `campaign_name_custom` extracts the bit of the URL that comes after "utm_campaign".  Tested for a few examples and appears to work, and is confirmed with stakeholders.  LGTM

PR #845, opened 2025-04-08
Update device-migration-accounts-toolbar-icon.toml

pissac17:


mikewilli (commented):
Did this cause a problem with the last analysis run? It doesn't look like analysis failed and I see results for this metric in Experimenter. If it's just a formatting change, I'd suggest adding `[ci rerun-skip]` to the PR title so we don't incur unnecessary costs by rerunning this.

PR #866, opened 2025-04-23
update desktop-release-show-relay-to-all-browsers.toml

irrationalagent:


jetstream/desktop-release-show-relay-to-all-browsers.toml
```diff
@@ -3,5 +3,11 @@
...
 description = "User exposed to logins dropdown on a login form"
-data_source = "glean_events_stream"
-select_expression = "event_category = 'form_autocomplete' AND event_name = 'show'"
+data_source = "events_form_autocomplete"
+select_expression = "event_name = 'show_logins'"
+
+[data_sources]
+[data_sources.events_form_autocomplete]
+from_expression = "(SELECT legacy_telemetry_client_id, experiments, DATE(submission_timestamp), event_name FROM firefox_desktop.events_stream WHERE event_category = 'form_autocomplete')"
+experiments_column_type = "none"
```
irrationalagent, line 12: I wasn't sure about this, I'm following the pattern from [here](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L2173)

mikewilli: Yep, makes sense to me. I'd just add the submission_date alias and then that should be good (`DATE(submission_timestamp) AS submission_date`). You could also consider splitting it into multiple lines for clarity, but that's just a subjective thing.

PR #867, opened 2025-04-23
change relay metrics and outcomes

irrationalagent:


jetstream/desktop-release-show-relay-to-all-browsers.toml
```diff
@@ -6,8 +6,162 @@ description = "User exposed to logins dropdown on a login form"
...
+[metrics.relay_integration_shown_relay_offer]
+friendly_name = "Relay Optin: Relay offer shown to client"
+description = "Number of times Relay offer is shown"
+select_expression = """
+      COALESCE(COUNTIF(
+         event_method = 'shown'
+         AND event_object = 'offer_relay'
+      ),0)
+"""
+data_source = "relay_integration_events"
```
mikewilli, line 19: You can skip the stuff that stays the same, it should still get picked up from the outcome since the metric definitions aren't changing. Not a big deal though if you'd rather be explicit here.

jetstream/desktop-release-show-relay-to-all-browsers.toml
```diff
@@ -6,8 +6,162 @@ description = "User exposed to logins dropdown on a login form"
...
 client_id_column = "legacy_telemetry_client_id"
+
+[data_sources.relay_integration_events]
+from_expression = """(
+      SELECT * FROM `mozdata.telemetry.events`
+      WHERE event_category = 'relay_integration'
+)"""
+experiments_column_type = "native"
+friendly_name = "Relay Integration Events"
+description = "Events Ping filtered to relay_integration category"
```
mikewilli, line 167: Likewise, this shouldn't be needed unless it changed (but fine to include if you'd like).

PR #868, opened 2025-04-24
feat: change default analysis unit support to both client and group

mikewilli:
Now that profile_group_id has been added to the vast majority of data sources, we can switch to both client and group ID as the default.

cc @jaredlockhart


scholtzan (approved):

scholtzan:
https://github.com/mozilla/metric-hub/blob/018fc7777e33667ead1a140522b215bdf7bfa1f7/.circleci/config.yml#L6 might need to be updated to 1.3.0

PR #874, opened 2025-04-28
CTR for Ads HNT

jaredsnyder:
Adds CTR for Ads on the Newtab Homepage


definitions/firefox_desktop.toml
```diff
@@ -1115,6 +1115,15 @@ description = """
 Count of impressions of Sponsored Pocket content.
 """
 
+[metrics.sponsored_pocket_ctr]
```
danielkberry, line 1118: For experiments this will compute the average over clients of the client-level CTRs. Is that desired?

jetstream/defaults/firefox_desktop.toml
```diff
@@ -12,6 +12,8 @@ weekly = [
   "search_count",
   "qualified_cumulative_days_of_use",
   "is_default_browser",
+  "sponsored_pocket_ctr",
```
danielkberry, line 15: So the intention is to calculate these metrics for every Firefox desktop experiment?

jetstream/defaults/firefox_desktop.toml
```diff
@@ -12,6 +12,8 @@ weekly = [
...
+  "sponsored_tile_ctr",
   "client_level_daily_active_users_v2",
 ]
```
danielkberry, line 19: These metrics need a statistic associated with them. Historically that's been defined in the (confusingly named) `/jetstream/definitions/firefox_desktop.toml` [file](https://github.com/mozilla/metric-hub/blob/main/jetstream/definitions/firefox_desktop.toml)

definitions/firefox_desktop.toml
jaredsnyder: No that is not ideal, is there any way to have it calculated at the day level?  Or is this just a fundamental part of how jetstream/etc works?

jetstream/defaults/firefox_desktop.toml
jaredsnyder: Thank you I was just trying to figure out what was going on there

jetstream/defaults/firefox_desktop.toml
jaredsnyder: It only really applies to newtab experiments so I guess not, changing now

definitions/firefox_desktop.toml
jaredsnyder: This might be a separate question by where do I set the analysis period?  Is it strictly a documentation thing?

danielkberry: It sounds like you'd want to make a change to make the metric a "ratio" metric where the numerator and denominator are summed independently. To do that you would define the numerator and denominator separately as metrics then define the metadata to make the ratio ([example)](https://github.com/mozilla/metric-hub/blob/main/definitions/firefox_desktop.toml#L3-L12). Then you give that ratio metric the `population_ratio` statistic and specific the numerator and denominator ([example](https://github.com/mozilla/metric-hub/blob/main/jetstream/outcomes/firefox_desktop/firefox_suggest.toml#L12-L14)). 

Analysis periods are explicitly specified for custom metrics or for defaults. For metrics in Outcomes, they're automatically set to `overall` and `weekly`. Does this absolutely have to be daily?

danielkberry: It'd be a little more costly to compute, querying data every day in addition to 1x/week. However, more urgently is the issue that outcomes are currently only weekly and overall. So this would have to be changed somehow, or this metric is added as a default to all experiments. 

Is this maybe something better included on a monitoring dashboard for all experiments than as a statistical output for some of them? I'm trying to understand the use case here

jaredsnyder: In that case I think weekly is fine

jaredsnyder: In that case I think weekly is fine, it's not so important we need to urgently add daily to outcomes or something

jetstream/outcomes/firefox_desktop/newtab_do_no_harm.toml
```diff
@@ -17,3 +17,23 @@ Guardrail metrics for the New Tab. Includes user experience and revenue-related
 [metrics.newtab_engagement.statistics.binomial]
 [metrics.newtab_any_searches.statistics.binomial]
 [metrics.newtab_gt4_searches.statistics.binomial]
+
+[metrics.sponsored_tile_clicks.statistics.sum]
```
danielkberry, line 21: the `sum` is kinda a weird statistic, it sums the client-level values into a total per branch, but doesn't have confidence intervals or anything. If you have any interest in looking at these middle metrics, you might want to also add one of the `_mean` statistics so that you get confidence intervals

danielkberry (approved):
Mostly LGTM. I left a comment about another statistic you could consider adding, but I won't block on it

PR #878, opened 2025-04-30
[ci rerun-skip] fix: linear model mean failing to fit, using bootstrap

mikewilli:


irrationalagent:
Since I also had the metrics defined [in the config](https://github.com/mozilla/metric-hub/blob/main/jetstream/desktop-release-show-relay-to-all-browsers.toml), do we need to change it there too? (or remove them?)

mikewilli:
Oops good catch!

irrationalagent (approved):

PR #880, opened 2025-04-30
DS-4224 updating metrics

YeonjooSmith:


jetstream/change-app-color-icon-android-for-existing-users-v2.toml
```diff
@@ -1,20 +1,26 @@
...
+overall = ["app_opened", "app_icon_used", "app_opens_per_client", "app_icon_uses_per_client"]
 
 [metrics.app_opened]
 select_expression = "CAST(SUM(IF(event.name = 'app_opened',1,0)) >0 AS INT)"
 data_source = "events"
 friendly_name = "app opened"
 description = "app opened"
 
+[metrics.app_opens_per_client]
+select_expression = "SUM(IF(event.name = 'app_opened',1,0))"
```
danielkberry, line 19: How should you handle a client who never opens the app... in that case they would never send an event. Right now, they would be represented as a `NULL` and wouldn't count toward the metric. Is this the intention? If you want them to count as a zero, you should wrap this in a `COALESCE(..., 0)`

danielkberry (approved):
Added a comment (it applies to both metrics) but it's not blocking

YeonjooSmith:
@danielkberry Would you mind reviewing the metric definitions that I added to count the number of app opened per client and the number of the app icon used? Also I think the statistical analysis I should use for the counting metrics is the bootstrap mean to check if the average count per sample is different from two group. Please let me know if it is appropriate method in this situation.

danielkberry:
@YeonjooSmith Yes I think looking at the average (bootstrap mean) is appropriate here. What do you think about my comment on appropriately counting users who never open the app?

jetstream/change-app-color-icon-android-for-existing-users-v2.toml
YeonjooSmith: Thank you for pointing this out. I am updating both new metric definitions.

YeonjooSmith:
@danielkberry Thank you for the review. I updated the metrics definition to output 0 when the user does not engage with the app opening or the app icon usage during the experiment.

PR #900, opened 2025-05-09
fix funnel metrics to use client_count statistic for accurate aggregates

jaredsnyder:
Was summing clients, which made aggregates in looker works.  Seems like this is the problem solved by the `client_count` statistic, so update to use that.


danielkberry (approved):
lgtm

PR #903, opened 2025-05-13
Create search-result-de-duplication.toml

cbellini:
custom metrics for first result deduplication experiment

### DIFF
--- a/jetstream/search-result-de-duplication.toml
+++ b/jetstream/search-result-de-duplication.toml
@@ -1,8 +1,10 @@
 [metrics]
-weekly = ["history_impressions", "bookmark_impressions", "search_suggestion_impressions",
-          "history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr", "search_suggestion_ctr"]
+weekly = ["history_impressions", "bookmark_impressions", "search_suggestion_impressions", 
+"history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr", 
+"search_suggestion_ctr", "history_impression_rate", "bookmark_impression_rate", "search_suggestion_impression_rate"]
 overall = ["history_impressions", "bookmark_impressions", "search_suggestion_impressions",
-           "history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr", "search_suggestion_ctr"]
+"history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr", 
+"search_suggestion_ctr", "history_impression_rate", "bookmark_impression_rate", "search_suggestion_impression_rate"]
 
 ## Impressions
 [metrics.history_impressions]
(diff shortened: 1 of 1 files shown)


shell1 (commented):
need to adapt the impression rate and CTR.  For CTR need to add the statistics after each one.

CTR is missing the 171 part from here https://github.com/mozilla/metric-hub/blob/d2bc37c87ec13cd4aa157d034dcf3fcd06e91c51/jetstream/fakespot-in-suggest.toml#L48-L53. 

example below for what history should have below the first part (which are right for all 3 - but all 3 missing the second stat part)...

[metrics.history_ctr.statistics.population_ratio]
numerator = "history_clicks"
denominator = "history_impressions"

jetstream/search-result-de-duplication.toml
```diff
@@ -0,0 +1,95 @@
...
+"search_suggestion_ctr", "history_impression_rate", "bookmark_impression_rate", "search_suggestion_impression_rate"]
+overall = ["history_impressions", "bookmark_impressions", "search_suggestion_impressions",
+"history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr", 
+"search_suggestion_ctr", "history_impression_rate", "bookmark_impression_rate", "search_suggestion_impression_rate"]
+
+## Impressions
+[metrics.history_impressions]
+select_expression = "COUNT(DISTINCT IF(is_terminal AND result.product_result_type = 'history', event_id, NULL))"
+data_source = "urlbar_events_unnested_results"
+[metrics.history_impressions.statistics.bootstrap_mean]
```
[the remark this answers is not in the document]
m-d-bowerman: This lets the automated analysis know what stats test to run for, in this case, the `history_impressions` metric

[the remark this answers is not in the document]
m-d-bowerman: You'll see a couple more around depending on the metric type/info wanted, like a `binomial` test for binary metrics or the `population_ratio` calculation

jetstream/search-result-de-duplication.toml
```diff
@@ -0,0 +1,95 @@
...
+[metrics.search_suggestion_ctr]
+depends_on = ["search_suggestion_clicks", "search_suggestion_impressions"]
+friendly_name = "Search Suggestion CTR"
+description = "Proportion of urlbar sessions where suggestion from the default search engine was clicked, out of all urlbar sessions where one was shown (not available on control)"
+
+[metrics.search_suggestion_ctr.statistics.population_ratio]
+numerator = "search_suggestion_clicks"
+denominator = "search_suggestion_impressions"
+
+## Impression Rate
```
shell1, line 69: the impression rate metrics depends on something called "urlbar_impressions" I didn't see it defined anywhere in the original reference file from "fakespot in suggest" https://github.com/mozilla/metric-hub/blob/main/jetstream/fakespot-in-suggest.toml

so i'm assuming that is just a know metric in metric hub and fine to use without something else it depends on.

cbellini: I think it is. I see `urlbar_pressions` in https://github.com/mozilla/metric-hub/blob/main/jetstream/yelp-suggestions-pilot-v3.toml.

m-d-bowerman: Strange, I don't see it in `firefox_desktop` definitions. I wonder if it was removed sometime; we'll need to define it either there or in this config.

jetstream/search-result-de-duplication.toml
```diff
@@ -0,0 +1,95 @@
+[metrics]
+weekly = ["history_impressions", "bookmark_impressions", "search_suggestion_impressions", 
+"history_clicks", "bookmark_clicks", "search_suggestion_clicks", "history_ctr", "bookmarks_ctr",
```
m-d-bowerman, line 3: There's a typo here causing the CI check to fail (`bookmarks_ctr` -> `bookmark_ctr`).

m-d-bowerman (changes requested):

jetstream/search-result-de-duplication.toml
shell1: Michael resolved here https://github.com/mozilla/metric-hub/pull/932 - as soon as that lands - this will work.

shell1 (approved):

PR #932, opened 2025-06-06
adds a urlbar_impressions metric

m-d-bowerman:
Add urlbar_impressions metric for search and suggest experiments.


shell1 (approved):
this looks like what we needed for the suggest metrics :)

PR #937, opened 2025-06-09
fix: try to make metrics query less complex

mikewilli:
In the cases where we're doing no-ops for their metrics, we change the data sources themselves to essentially no-ops as well (date filter to a single day outside of the experiment window, only select client_id and submission_date).

I also added a channel filter to the `newtab_visits_topsite_tile_interactions` data source because it's a cluster field. I don't expect a big difference, but it did cut down a tiny bit on the expected data to be processed. If we need to optimize further, we should eliminate that data_source altogether.

### DIFF
--- a/jetstream/first-time-wallpaper-newtab.toml
+++ b/jetstream/first-time-wallpaper-newtab.toml
@@ -15,3 +15,45 @@ select_expression = 'SUM(0)'
 select_expression = 'SUM(0)'
 [metrics.new_tab_enabled_on_new_tabs]
 select_expression = 'SUM(0)'
+
+
+[data_sources.main]
+from_expression = """(
+    SELECT
+        client_id,
+        DATE(submission_timestamp) AS submission_date
+    FROM `moz-fx-data-shared-prod.telemetry_stable.main_v5`
+    WHERE DATE(submission_timestamp) = '2025-03-04'
+)"""
+experiments_column_type = "none"
+friendly_name = "Main no-op"
+description = "Main ping table without meaningful metrics"
+
+[data_sources.as_sessions]
+from_expression = """(
+    SELECT
+        client_id,
+        DATE(submission_timestamp) AS submission_date
+    FROM mozdata.activity_stream.sessions
+        WHERE DATE(submission_timestamp) = '2025-03-04'
+    )"""
+experiments_column_type = "none"
+friendly_name = "Activity Stream Sessions no-op"
+description = "Activity Stream Sessions without meaningful metrics"
+
+[data_sources.newtab_visits_topsite_tile_interactions]
+from_expression = """(
+    SELECT
+        e.* EXCEPT (topsite_tile_interactions),
+        topsite_tile_interactions
+    FROM
+        `moz-fx-data-shared-prod.telemetry.newtab_visits` e
+    CROSS JOIN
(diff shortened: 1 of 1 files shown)


jetstream/first-time-wallpaper-newtab.toml
```diff
@@ -15,3 +15,45 @@ select_expression = 'SUM(0)'
...
 select_expression = 'SUM(0)'
+
+
+[data_sources.main]
+from_expression = """(
+    SELECT
+        client_id,
+        DATE(submission_timestamp) AS submission_date
+    FROM `moz-fx-data-shared-prod.telemetry_stable.main_v5`
+    WHERE DATE(submission_timestamp) = '2025-03-04'
```
danielkberry, line 26: Is the single date intended here so that it won't scan too much data?

jetstream/first-time-wallpaper-newtab.toml
```diff
@@ -15,3 +15,45 @@ select_expression = 'SUM(0)'
...
+        client_id,
+        DATE(submission_timestamp) AS submission_date
+    FROM mozdata.activity_stream.sessions
+        WHERE DATE(submission_timestamp) = '2025-03-04'
+    )"""
+experiments_column_type = "none"
+friendly_name = "Activity Stream Sessions no-op"
+description = "Activity Stream Sessions without meaningful metrics"
+
+[data_sources.newtab_visits_topsite_tile_interactions]
```
danielkberry, line 44: This table isn't a no-op but you added a channel filter, is that right?

jetstream/first-time-wallpaper-newtab.toml
```diff
@@ -15,3 +15,45 @@ select_expression = 'SUM(0)'
...
+from_expression = """(
+    SELECT
+        e.* EXCEPT (topsite_tile_interactions),
+        topsite_tile_interactions
+    FROM
+        `moz-fx-data-shared-prod.telemetry.newtab_visits` e
```
danielkberry, line 50: for the non-no-op (the op) table, would it help if we added filters on submission date or are those already covered?

danielkberry (approved):
r+ w/comments

jetstream/first-time-wallpaper-newtab.toml
mikewilli: Yes, that's correct. That one is still used for actual metrics.

jetstream/first-time-wallpaper-newtab.toml
mikewilli: Those are already covered (at least per the BQ console data processing estimates). I guess it couldn't hurt to add them here though?

jetstream/first-time-wallpaper-newtab.toml
mikewilli: Yep, maybe it would be better to just do a filter that won't get any data at all, but this was easiest and I think will do the trick. Open to a better filter though!

ilanasegall: can we just do "select 0"?

mikewilli: It needs the `client_id` and `submission_date` columns so that the `JOIN` conditions don't cause an error.

PR #941, opened 2025-06-11
Move webcompat metrics to events_stream

bvandersloot-mozilla:
it was too slow before!

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -2226,6 +2226,7 @@ from_expression = """(
         DATE(submission_timestamp) AS submission_date,
         experiments,
         event_name,
+        event_extra,
     FROM `mozdata.firefox_desktop.events_stream`
     WHERE
         event_category = 'webcompatreporting'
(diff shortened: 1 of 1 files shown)


definitions/firefox_desktop.toml
```diff
@@ -2238,14 +2237,13 @@ description = "Webcompat 'Report Broken Site' pings"
...
 from_expression = """(
     SELECT
-        metrics.uuid.legacy_telemetry_client_id AS client_id,
+        client_id,
         DATE(submission_timestamp) AS submission_date,
-        ping_info.experiments AS experiments,
-        event.name AS event_name,
+        experiments,
+        event_name,
     FROM `mozdata.firefox_desktop.events`
```
mikewilli, line 2244:
```suggestion
    FROM `mozdata.firefox_desktop.events_stream`
```

mikewilli (approved):
Thanks for bearing with me through the CI validation issues. Validating locally succeeds now so I think we're good to go!

PR #944, opened 2025-06-13
ETP and webcompat reporter statistics should be binomial

bvandersloot-mozilla:
Bootstrap means don't work because the data is driven by outliers. We want to look for a surge in outliers, not ignore them!


mikewilli (changes requested):
Thanks! I think this would work, but see my comment for a suggestion. And sorry that my "example" was not even an example, but glad you were able to figure it out.

mikewilli (approved):

PR #961, opened 2025-06-20
add ad_metrics_daily to ads definitions

mozza-wins:
Adding these metrics in for ad_metrics_daily to use them in looker, and maybe longterm in Jetstream and Opmon too. This table seems long term and broadsweeping enough to be in definitions instead of Looker


chelseybeck (approved):
lgtm

PR #991, opened 2025-06-30
add countries join to ad-metrics-daily

mozza-wins:
Adds a country join to ad-metrics-daily to get region


sergiosonline (approved):
Looks good. Thank you.

PR #997, opened 2025-07-03
Fix ratio statistics naming and remove unused metrics

mozza-wins:
The problem with these statistics was that we weren't able to surface the metric before, the `depends_on` functionality was not yet live so this is a workaround to surface the metrics so we can create ratio statistics with them. The reason to go this route and not just name the ratio after the numerator is because we have some ratios that use the same numerator and would not be able to share a name. Some naming conventions though are actually commonly used, like clicks.ratio is common for CTR, so those seemed ok to keep. 

My general approach was to only use SUM(1) in the looker layer metric definitions because it's solely for looker purposes. 

For the future, if `depends_on` does become live, we can change all SUM(1) instances back to `depends_on`

### DIFF
--- a/looker/definitions/firefox_desktop.toml
+++ b/looker/definitions/firefox_desktop.toml
@@ -382,18 +382,13 @@ description = "Count of clients with others engagement"
 [metrics.sponsored_impressions.statistics.client_count]
 
 [metrics.sponsored_tile_clicks.statistics.ratio]
-numerator = "sponsored_tile_clicks.sum'
-denominator = "sponsored_tile_impressions.sum'
+numerator = "sponsored_tile_clicks.sum"
+denominator = "sponsored_tile_impressions.sum"
 
 [metrics.sponsored_pocket_clicks.statistics.ratio]
-numerator = "sponsored_pocket_clicks.sum'
-denominator = "sponsored_pocket_impressions.sum'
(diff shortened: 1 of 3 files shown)


jaredsnyder (approved):
lgtm

looker/definitions/firefox_desktop.toml
```diff
@@ -381,52 +381,50 @@ description = "Count of clients with others engagement"
...
-denominator='sponsored_tile_impressions.sum'
+[metrics.sponsored_tile_clicks.statistics.ratio]
+numerator = "sponsored_tile_clicks.sum"
+denominator = "sponsored_tile_impressions.sum"
 
-[metrics.sponsored_pocket_ctr.statistics.ratio]
-numerator='sponsored_pocket_clicks.sum'
-denominator='sponsored_pocket_impressions.sum'
+[metrics.sponsored_pocket_clicks.statistics.ratio]
+numerator = "sponsored_pocket_clicks.sum"
```
scholtzan, line 389: `[metrics.sponsored_pocket_clicks.statistics.sum]`

needs to be defined in that file

mozza-wins: thanks @scholtzan, I'll update this in a following PR, wondering if maybe we should document all the learnings we've had around ratios in the docs? (Unless it's already written somewhere and I missed it)

PR #1027, opened 2025-07-23
remove analysis units where not needed

mozza-wins:
Removing all the `analysis_units` params from metrics, turns out they're not needed there.


danielkberry:
Is the `analysis_units` field on all of the data sources for these metrics? In theory our CI should check this, but I want to be double sure

danielkberry (approved):

PR #1041, opened 2025-07-30
[ci rerun-skip] Adds tagged_search_count to default overall metrics

shell1:
"tagged_search_count", was missing from Overall metrics so not being calculated


mikewilli (approved):
LGTM. Seems like this was just an accidental omission, since it's present in other app defaults.

PR #1055, opened 2025-08-06
Create client-side-relevance-ranking-for-enhanced-firefox-suggest-pha…

shell1:
…se-2.toml


mikewilli:
This is trying to run validation on a file that doesn't exist, which makes me think it's just out of date. Try the "Update branch" button and see if it still fails after that.

PR #1073, opened 2025-08-14
custom configs for traffic-impact-studies

mozza-wins:
Creates custom configs for these experiments ([traffic-impact-study-1](https://experimenter.services.mozilla.com/nimbus/traffic-impact-study-1), [2](https://experimenter.services.mozilla.com/nimbus/traffic-impact-study-2/summary/), [3](https://experimenter.services.mozilla.com/nimbus/traffic-impact-study-3/summary/)) so we can rerun the results on a smaller subset of relevant metrics

### DIFF
--- a/jetstream/traffic-impact-study-1.toml
+++ b/jetstream/traffic-impact-study-1.toml
@@ -1,0 +1,4 @@
+[metrics]
+weekly = ['sponsored_tile_impressions', 'sponsored_tile_clicks', 'sponsored_tile_ctr', 'sponsored_tiles_disabled', 'sponsored_tiles_dismissals']
+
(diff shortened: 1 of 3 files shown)


mikewilli (changes requested):
Unfortunately, this won't work the way you're expecting. The custom config takes the existing config and merges itself into that, so the weekly/overall metrics lists will be a combination of what you have here and any other metrics configured (so you'll end up with the same lists after jetstream de-duplicates the lists).

And also unfortunately, the way to skip over metrics is a bit annoying at the moment. What you'll want to do is create a no-op datasource, and then use it for each of the metrics you **don't want** (should look something like [this](https://github.com/mozilla/metric-hub/blob/main/jetstream/intreaction-frecency-enhancement.toml#L68-L82) metric and data_source). You can ignore the ones you want to keep in the custom configs.

mikewilli (approved):
Looks good. Let's do `[ci rerun-skip]` (in the PR title) on this one and I can rerun it manually with just overall.

PR #1074, opened 2025-08-14
create new outcome for tiles incrementality tests

mozza-wins:
For future incrementality tile tests, we'll use this set of metrics that are relevant to the product.

### DIFF
--- a/jetstream/outcomes/firefox_desktop/incrementality_sponsored_tiles.toml
+++ b/jetstream/outcomes/firefox_desktop/incrementality_sponsored_tiles.toml
@@ -1,5 +1,5 @@
-friendly_name = "Sponsored Tiles"
-description = "Interaction metrics for sponsored tiles."
+friendly_name = "Incrementality Sponsored Tiles"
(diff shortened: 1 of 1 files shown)


jetstream/outcomes/firefox_desktop/incrementality_sponsored_tiles.toml
```diff
@@ -0,0 +1,18 @@
+friendly_name = "Sponsored Tiles"
```
mikewilli, line 1: This should be a unique name (not strictly required, but will be confusing if it's not).

mikewilli (changes requested):
Syntax looks fine, just one change requested.

Also worth considering whether this should be a new outcome. It looks very similar to the existing Sponsored Tiles outcome, would it make more sense to edit that one?

mikewilli (approved):

mikewilli:
Gotcha, as long as you've considered it and think it's a worthwhile distinction then I'm happy. Thanks! LGTM

PR #1102, opened 2025-08-26
Add captcha detection JetStream Outcome

FKLC:
Hello,

We would like to add our [captcha detection](https://firefox-source-docs.mozilla.org/toolkit/components/captchadetection/captcha-detection/index.html) metrics to the experimenter. I think this covers it, but I'll be able to fix issues with it.

Do note that this depends on https://phabricator.services.mozilla.com/D262592.

Thank you!


FKLC:
Just to be clear, I'm no longer a Mozilla employee. If there are legal or other implications, the next best contact for this whole thing is Tom Ritter.

PR #1205, opened 2025-11-02
Sponsored tiles outcome metrics

soGaussian:
Migrated the sponsored tiles outcome metrics to use event_stream instead of events


definitions/fenix.toml
```diff
@@ -596,6 +596,22 @@ description = "Events Ping"
...
+[data_sources.events_stream]
+from_expression = """(
+    SELECT
+        p.*,
+        DATE(p.submission_timestamp) AS submission_date
+    FROM
+        `moz-fx-data-shared-prod.{dataset}.events_stream` p
+)"""
+client_id_column = "client_id"
+experiments_column_type = "glean"
```
mikewilli:
```suggestion
experiments_column_type = "events_stream"
```

definitions/fenix.toml
```diff
@@ -596,6 +596,22 @@ description = "Events Ping"
...
+default_dataset = "org_mozilla_firefox"
```
mikewilli: Should this just be `fenix`? I always get confused about the various fenix datasets..

definitions/fenix.toml
```diff
@@ -596,6 +596,22 @@ description = "Events Ping"
 build_id_column = "REPLACE(CAST(DATE(mozfun.norm.fenix_build_to_datetime(client_info.app_build)) AS STRING), '-', '')"
 
 
+[data_sources.events_stream]
+from_expression = """(
+    SELECT
+        p.*,
+        DATE(p.submission_timestamp) AS submission_date
```
mikewilli: Not a big deal, but it's better to select what you need if possible. I think you should be ok with just this:
```suggestion
        legacy_telemetry_client_id AS client_id,
        profile_group_id,
        experiments,
        event_category,
        event_name,
        event_extra,
        DATE(p.submission_timestamp) AS submission_date
```
but up to you.

mikewilli (dismissed):
Good start, left you some suggestions on a few things to change.

scholtzan (approved):

PR #1213, opened 2025-11-07
AD-531 Remove outdated SPOCs looker view

curtismorales:
Remove the `consolidated_ads_spocs` generated Looker view. This view is not currently accurate or in use, and the upstream data source has been deprecated.


jaredsnyder (approved):
nice

PR #1237, opened 2025-11-17
remove newtab_clients_daily from looker definitions

jaredsnyder:
newtab_clients_daily and related metrics are not currently being used, and can be removed.


ilanasegall (commented):
R+

PR #1239, opened 2025-11-18
[ci rerun-skip] replace pocket interactions dataset with newtab visits and update exposure_signal

jaredsnyder:
The pocket interactions dataset is huge and makes experiments fragile.  This replaces it with the much smaller newtab_visits datatset and updates the exposure signal accordingly


land-edi (approved):
Let's :shipit:

PR #1252, opened 2025-12-02
Incrementality metric refactor

jaredsnyder:
Remove metric based on a broken table

Want to rerun `traffic-impact-study-9` so purposefully left off [ci skip rerun] but need to verify that makes sense (@mikewilli )


mikewilli (approved):
This won't rerun any experiments since it only affects an outcome. You'd need to change the desired experiments' custom config files (of the form `jetstream/experiment-slug.toml`) for them to be picked up and rerun.

That said, you could also use the `jetstream_rerun` DAG for cases like this.

PR #1256, opened 2025-12-03
remove deprecated wallpaper and widget  metrics

jaredsnyder:
Removing wallpaper and widget metrics that have not been used in some time and will not be supported in the new version of newtab_clients_daily


ilanasegall (commented):
approved as we are not looking at these metrics at the experiment level at this time

PR #1258, opened 2025-12-04
Overwrite irrelevant  outcome metrics

YeonjooSmith:
overwriting outcome metrics to resolve the errors discussed in this thread: https://mozilla.slack.com/archives/CF94YGE03/p1764707293909879?thread_ts=1764087234.436709&cid=CF94YGE03

### DIFF
--- a/jetstream/firefox-vpn-for-windows-10-users.toml
+++ b/jetstream/firefox-vpn-for-windows-10-users.toml
@@ -104,21 +104,6 @@ data_source = "noop_source"
 
 # overwrite all metrics in Networking Performance outcome: https://github.com/mozilla/metric-hub/blob/main/jetstream/outcomes/firefox_desktop/networking_performance.toml
 
-[metrics.http_page_tls_handshake]
-select_expression = 'SUM(0)'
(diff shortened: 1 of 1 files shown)


jetstream/firefox-vpn-for-windows-10-users.toml
```diff
@@ -1,2 +1,592 @@
...
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+
+[metrics.dns_lookup_time.statistics.kernel_density_estimate]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+
+[metrics.dns_lookup_time.statistics.empirical_cdf]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
```
mikewilli: The `select_expression` and `data_source` parameters are on the metric, not on the statistic. You also don't need to worry about re-defining the statistics at all for the purpose of skipping metrics, so what you end up with should look like this:

```suggestion
[metrics.dns_lookup_time]
select_expression = 'SUM(0)'
data_source = "noop_source"
```

PR #1273, opened 2025-12-12
add config for frecency experiment

jaredsnyder:
Doc: https://docs.google.com/document/d/1gJfx8ITCwBOuy2BiRux_fATgAYW1UxZicAbIcYFdNJc/edit?tab=t.erl2a62ns55c


mikewilli (changes requested):
I left one specific comment on a change I know you'll need. There are some other issues here that you'll likely be able to find yourself and fix once jetstream preview is fixed in https://github.com/mozilla/jetstream/pull/2936, but the problem is with the `total_tile_third*` and `sponsored_tile*` metrics -- metric-config-parser is saying some of them are missing statistics.

mikewilli:
For posterity:

As far as @jaredsnyder and I could tell, the order in which metrics were defined in the daily/weekly/overall lists made a difference in whether the config resolved properly (see [this commit](https://github.com/mozilla/metric-hub/pull/1273/changes/8a2f49762414460fd833c09d8bbeaf2b021a96ea)), at least for the metrics that depend on others. This appears to be because the parser just pulls in and resolves the metrics in the order they're defined, and so if the downstream metric is in the list before its dependencies, those dependent metrics don't exist as far as the parser knows.

I also filed this ticket https://github.com/mozilla/metric-hub/issues/1295

PR #1276, opened 2025-12-15
feat: delete definitions related to focus_android, focus_ios, klar_android, and klar_ios

kik-kik:
# feat: delete definitions related to focus_android, focus_ios, klar_android, and klar_ios

The DAU related metrics should still be available via mobile_active_users_aggregates_view data source.


scholtzan:
The `definitions/` and `opmon/` files should not be deleted, only what's in `looker/`. `definitions/` and `opmon/` files are used for experiments and opmon projects.

PR #1282, opened 2025-12-16
Update metrics from pocket to content

scholtzan:
The metric names have changed in https://github.com/mozilla/metric-hub/commit/d03b23a11a7592057cc556c41468b71dee1cc8b1

Not sure if this rollout is still relevant? If we need historical data, then a backfill needs to be triggered via https://workflow.telemetry.mozilla.org/dags/operational_monitoring_backfill/grid?search=operational_monitoring_backfill


jaredsnyder:
I am still using this as a test case for using opmon for rollouts, though it has gotten quite stale 😅

PR #1385, opened 2026-04-07
Custom config for splitview feature notification exp

land-edi:
Create custom config by creating enrollments, custom metrics and segments

### DIFF
--- a/jetstream/os-notification-split-view-release-lapsed-users.toml
+++ b/jetstream/os-notification-split-view-release-lapsed-users.toml
@@ -254,7 +254,8 @@ from_expression = """(
     ),
     week_2_retention AS (
         SELECT
-            bse.*,
+            bse.* EXCEPT(event_extra),
+            ANY_VALUE(event_extra) AS event_extra,
             LOGICAL_OR(COALESCE(wke.used_splitview, FALSE)) AS week_2_retained
         FROM
             feature_usage AS bse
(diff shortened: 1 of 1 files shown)


jetstream/os-notification-split-view-release-lapsed-users.toml
```diff
@@ -0,0 +1,283 @@
...
+            client_info.client_id AS client_id,
+            ping_info,
+            TRUE AS used_splitview
+        FROM 
+            `moz-fx-data-shared-prod.firefox_desktop.events` e,
+            UNNEST(e.events) event
+        WHERE 
+            event.category = 'splitview'
+            AND event.name = 'start'
+            AND DATE(submission_timestamp) BETWEEN '2026-03-26' AND '2026-04-30'
```
mikewilli, line 251: `events_stream` is preferable over `events`, unless there's a particular reason it won't work.

mikewilli (commented):
Looks good overall, seems like a pretty typical change for the background update enrollments query. Please change the one data source to use `events_stream` though, it should be much more efficient than having to UNNEST the events directly.

jetstream/os-notification-split-view-release-lapsed-users.toml
land-edi: @mikewilli I originally used events_stream, but got an error as the Jetstream validation CI was looking for the `ping_info` column in that dataset. I did not see ping_info in events_stream, so I switched to events. Is `ping_info` available in events_stream?

PR #1391, opened 2026-04-16
small fix for mcp

mozza-wins:
small thing I found when using the mcp json


chelseyklein (approved):
this is right according to docs, but i haven't tested personally

mozza-wins:
thanks Chelsey, i tested it on mine :^)

PR #1396, opened 2026-04-20
feat(nimbus): Nimbus feature monitoring

### LINKED ISSUE #15387: Move configuration files to experimenter github repo
feature monitoring configs like [https://github.com/mozilla/bigquery-etl/blob/.../sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml](https://github.com/mozilla/bigquery-etl/blob/5550aeed344dd635d0de26f5f62d1d69d9a5e36d/sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml) should live in the metric hub repo. The config format should be changed from YAML to TOML, and the source table property is_event_stream should be changed to type so that we explicitly support specific table types, e.g. metrics (currently defined by is_event_stream: false) event_stream, and clients daily (not specifically supported yet). The bigquery-etl generator python should then use metric-config-parser to get the configs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6732)

yashikakhurana:
Nimbus feature monitoring TOML configs
## What

Implements EXP-6732 — moves Nimbus feature monitoring configs from bigquery-etl into metric-hub.

### New files

**`nimbus_feature_monitoring/firefox_desktop.toml`**
- Converts `firefox_desktop.yaml` from bigquery-etl to TOML
- Replaces `is_events_stream: true` with `type = "event_stream"`
- Source tables now have explicit `type` of `metrics`, `event_stream`, or `clients_daily`

**`lib/metric-config-parser/metric_config_parser/nimbus_feature_monitoring.py`**
- New `NimbusFeatureMonitoringSpec` class following the same `@attr.s` / `toml.load` patterns as existing specs (`AnalysisSpec`, `MonitoringSpec`)

- bigquery-etl PR: [(9229)](https://github.com/mozilla/bigquery-etl/pull/9229)

closes [#5387 ](https://github.com/mozilla/experimenter/issues/15387)


jaredlockhart (changes requested):
Sweet! So where is the original YAML config?

https://github.com/mozilla/bigquery-etl/blob/d64e5cfd4511dac62c077fed62faa21b121da322/sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml

Are we gonna auto convert that into the metric hub format?

yashikakhurana:
> Sweet! So where is the original YAML config?
> 
> https://github.com/mozilla/bigquery-etl/blob/d64e5cfd4511dac62c077fed62faa21b121da322/sql_generators/nimbus_feature_monitoring/templates/firefox_desktop.yaml
> 
> Are we gonna auto convert that into the metric hub format?

@jaredlockhart this is the equivalent file here https://github.com/mozilla/metric-hub/pull/1396/changes#diff-a2117b2b7378afbdc0a3dfd7bfb22489947360cc3c80f030a45cdf61fbb6e453

nimbus_feature_monitoring/firefox_desktop.toml
```diff
@@ -0,0 +1,114 @@
...
+
+[source_tables.firefox_desktop_metrics.dimensions.normalized_channel]
+
+[source_tables.firefox_desktop_metrics.dimensions.normalized_os]
+
+[source_tables.firefox_desktop_metrics.dimensions.normalized_os_version]
+
+[source_tables.firefox_desktop_metrics.dimensions.normalized_country_code]
+value_in = ["CA", "DE", "FR", "GB", "US"]
+default_value = "OTHER"
```
mikewilli: I don't know if we like this better, but here's another option:


```suggestion
[source_tables.firefox_desktop_metrics.dimensions]
app_display_version = { field = "client_info.app_display_version" }
app_version_major = {}
app_version_minor = {}
app_version_patch = {}
locale = { field = "client_info.locale" }
normalized_channel = {}
normalized_country_code = { value_in = ["CA", "DE", "FR", "GB", "US"], default_value = "OTHER" }
normalized_os = {}
normalized_os_version = {}
```

nimbus_feature_monitoring/firefox_desktop.toml
```diff
@@ -0,0 +1,114 @@
+dataset = "firefox_desktop"
+
+# Source tables define the BigQuery tables and their schema used for feature monitoring.
+# Each source table specifies:
+#   - analysis_unit_id: the field used to identify analysis units (e.g. client/profile IDs)
+#   - table_name: the BigQuery table name within the dataset
+#   - type: one of "metrics", "event_stream", or "clients_daily"
+#   - dimensions: optional per-table dimensions to group metrics by
+
+[source_tables.firefox_desktop_metrics]
```
mikewilli: Should we say `data_sources` instead of `source_tables` to make this more consistent with naming elsewhere? Or is there a good reason to have a unique name here?

nimbus_feature_monitoring/firefox_desktop.toml
```diff
@@ -0,0 +1,114 @@
+dataset = "firefox_desktop"
```
mikewilli: suggestion(blocking): I think this should be implied from the filename, no?

jaredlockhart (approved):
Awesome thnx @yashikakhurana looks right to me but good to get @mikewilli to look as well 🙏

nimbus_feature_monitoring/firefox_desktop.toml
mikewilli: It seems like we're being quite redundant with all the "firefox_desktop" repetition throughout here. I guess just `source_tables.metrics` might be too generic though?

nimbus_feature_monitoring/firefox_desktop.toml
```diff
@@ -0,0 +1,114 @@
...
+
+# Source tables define the BigQuery tables and their schema used for feature monitoring.
+# Each source table specifies:
+#   - analysis_unit_id: the field used to identify analysis units (e.g. client/profile IDs)
+#   - table_name: the BigQuery table name within the dataset
+#   - type: one of "metrics", "event_stream", or "clients_daily"
+#   - dimensions: optional per-table dimensions to group metrics by
+
```
mikewilli: We have `example_config.toml.example` files in some of the folders to provide unused samples of what is expected. Could be useful here too? Example: https://github.com/mozilla/metric-hub/blob/main/jetstream/example_config.toml.example

nimbus_feature_monitoring/firefox_desktop.toml
```diff
@@ -0,0 +1,114 @@
...
+analysis_unit_id = "metrics.uuid.legacy_telemetry_profile_group_id"
+table_name = "newtab"
+type = "metrics"
+
+# Features define which metrics to collect from which source tables.
+# Each feature maps to metrics grouped by source table and data type.
+# For boolean, quantity, and labeled_* types: keys are metric names.
+# For event type: keys are categories, then event names.
+
+["features"."address-autofill-feature"."metrics_by_source"."firefox_desktop_metrics".boolean]
```
mikewilli: Hm, kind of annoying that these need to be quoted because the metric is hyphenated (at least I assume that's why). Would it be better to standardize the keys on snake_case and then have a separate parameter to specify the metric_name if it doesn't match?

nimbus_feature_monitoring/firefox_desktop.toml
mikewilli: This might match better with some of the syntax below for the features.

lib/metric-config-parser/metric_config_parser/nimbus_feature_monitoring.py
```diff
@@ -0,0 +1,93 @@
+"""Spec for Nimbus feature monitoring TOML configs.
+
+Config files live in the ``nimbus_feature_monitoring/`` directory of a metric-hub
+repository checkout, with one TOML file per application (e.g. ``firefox_desktop.toml``).
+
+Each file defines:
+- ``dataset``: the BigQuery dataset name for the application
+- ``source_tables``: BigQuery tables to query, each with a ``type`` of
+  ``"metrics"``, ``"event_stream"``, or ``"clients_daily"``
+- ``features``: Nimbus features and the metrics to collect per source table
```
mikewilli: suggestion(blocking): I'd suggest we put this in a README in the TOML folder as well.

lib/metric-config-parser/metric_config_parser/nimbus_feature_monitoring.py
```diff
@@ -0,0 +1,93 @@
...
+    To load all configs from a metric-hub repo checkout::
+
+        specs = NimbusFeatureMonitoringSpec.configs_from_repo(Path("/path/to/metric-hub"))
+        for app_name, spec in specs:
+            ...
+    """
+
+    dataset: str
+    source_tables: dict[str, Any] = attr.Factory(dict)
+    features: dict[str, Any] = attr.Factory(dict)
```
mikewilli: suggestion(blocking): I'd suggest defining a class for SourceTable (or whatever we call it) and Feature so that the expected fields are defined.

mikewilli (commented):
I left a bunch of thoughts, most of them are intended as discussion points as opposed to blocking critiques.

One thing I'm not sure of in general is what to call this. `nimbus_feature_monitoring` is fine, but seems rather verbose. Maybe just `feature_monitoring`, or something abbreviated like featmon to match opmon? Or is there another cloud-adjacent term we could use (a la jetstream). What do you think?

mikewilli:
You'll also need to bump the metric-config-parser version in its pyproject.toml so it gets published.

nimbus_feature_monitoring/firefox_desktop.toml
jaredlockhart: Oh yeah that looks better.  Interesting somethings that were expressed nicely in YAML are awkward in TOML so yeah this definitely looks better to me.

nimbus_feature_monitoring/firefox_desktop.toml
jaredlockhart: Ya consistency is good, good idea 👍

nimbus_feature_monitoring/firefox_desktop.toml
jaredlockhart: Oh we could do that, we do that in other places that's not a bad idea.

nimbus_feature_monitoring/firefox_desktop.toml
jaredlockhart: Well I think in the other structure it referred back to the 'firefox_desktop' identifier earlier? Not sure if that's structural here or just convention?

nimbus_feature_monitoring/firefox_desktop.toml
jaredlockhart: Oh yeah it's becuase it's the feature name.  Yeah I agree it's annoying to have to quote here but I don't think it's feasible to go and rename all teh features in the actual feature manifests, and I definitely don't wanna do something where we transform the slugs into snake here and then they don't match what's in the manifest.  Probably just gotta suffer the quotes here 🙏 😭 "

featmon/README.md
mikewilli, line 1: 🎉

mikewilli (approved):
Looks great, thanks! I have some ideas for how we can extend this in the future, but not necessary for this first pass. We can discuss those later once we have this working.

nimbus_feature_monitoring/firefox_desktop.toml
mikewilli: Yea in both versions (YAML and TOML) this is just convention. I definitely don't feel strongly about this, just seemed like we were saying "firefox_desktop" everywhere and thought it might be nice to be more concise.

nimbus_feature_monitoring/firefox_desktop.toml
mikewilli: Yea, fair enough. The option I was suggesting is that we could standardize on snake case for the metric-hub identifier, and then have the actual feature identifier be specified separately as a parameter, like:
```toml
[features.i_am_a_feature]
feature_id = "i-am-a-feature"
```

And like most of my other suggestions, I'm not sure which one I like more. Just presenting an alternative.

Sounds like you'd prefer the previous structure, and we just deal with the quotes in the `[features."i-am-a-feature"]` part?

yashikakhurana: oh I think I missed this comment before merging, any preference over this?

mikewilli: I don't have a real preference either way, I can see the pros/cons of each approach, so whichever one you merged is fine with me.

PR #1406, opened 2026-04-22
Read from local repo

yashikakhurana:
Part of [EXP-6732](https://mozilla-hub.atlassian.net/browse/EXP-6732) — moving Nimbus feature monitoring configs from hardcoded bigquery-etl YAML files into metric-hub TOML.

This PR makes featmon configs a first-class config type in `ConfigCollection`, so they load the same way as jetstream/opmon configs and can be accessed via `ConfigCollection.from_github_repo().featmon_configs.`

What's in here
-` FeatmonConfig `class — a slim wrapper around `FeatmonSpec (slug + spec), `following the same pattern as Outcome. Plugs into `entity_from_path() `and validate.py so featmon files get validated in CI like everything else.
- `ConfigCollection.featmon_configs` — `a list[FeatmonConfig] `populated automatically when you load from GitHub or a local repo. The old` FeatmonSpec.configs_from_repo() `method is removed since `ConfigCollection `handles this now.
- Two new MCP tools — `list_monitored_features` and` get_monitored_feature`, so AI assistants can discover and inspect featmon configs.
- Tests — TestFeatmonConfig covering the new class, plus 4 MCP handler tests.
Notes
- The featmon check in `entity_from_path()` runs before `validate_config_settings()` — featmon TOMLs use data_sources/features keys that would otherwise be rejected by the jetstream/opmon schema checker.
- The companion bigquery-etl PR updates the SQL generator to read from `ConfigCollection` instead of the local YAML templates.
- Bumps version to 2026.4.3.

Relates to EXP-6732
https://github.com/mozilla/experimenter/issues/15387
https://github.com/mozilla/experimenter/issues/15428

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/config.py
+++ b/lib/metric-config-parser/metric_config_parser/config.py
@@ -214,17 +214,40 @@ class DefinitionConfig(Config):
         analysis_spec.resolve(dummy_experiment, configs)
 
 
+@attr.s(auto_attribs=True)
+class FeatmonConfig:
+    """Represents a feature monitoring config file for a single application."""
+
+    slug: str  # app name inferred from file stem (e.g. "firefox_desktop")
+    spec: FeatmonSpec
+
+    def validate(self, configs: "ConfigCollection", _experiment=None) -> None:
+        # Structural validation (e.g. valid source types) is already performed by
+        # FeatmonSpec.__attrs_post_init__ at parse time. No cross-config validation
+        # is needed here, so this stub intentionally does nothing.
+        pass
+
+
 def entity_from_path(
     path: Path, is_private: bool = False
-) -> Config | Outcome | DefaultConfig | DefinitionConfig | FunctionsSpec:
+) -> Config | Outcome | DefaultConfig | DefinitionConfig | FunctionsSpec | FeatmonConfig:
     is_outcome = path.parent.parent.name == OUTCOMES_DIR
     is_default_config = path.parent.name == DEFAULTS_DIR
     is_definition_config = path.parent.name == DEFINITIONS_DIR
+    is_featmon = path.parent.name == FEATMON_DIR
     slug = path.stem
 
-    validate_config_settings(path)
-
     config_dict = toml.loads(path.read_text())
+
+    # featmon TOMLs have their own schema (data_sources/features) that is not
+    # compatible with validate_config_settings, which checks jetstream/opmon keys
+    if is_featmon:
+        return FeatmonConfig(
+            slug=slug,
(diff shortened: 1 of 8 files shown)


lib/metric-config-parser/metric_config_parser/config.py
```diff
@@ -314,6 +315,7 @@ class ConfigCollection:
     functions: FunctionsSpec | None = None
     repos: list[Repository] = attr.Factory(list)  # repos configs were loaded from
     is_private: bool = False
+    featmon_configs: dict[str, FeatmonSpec] = attr.Factory(dict)
```
mikewilli: Can we stick to the established pattern here and use a `list[FeatmonConfig]`? Should look similar to the `Outcome` class on line 150, except that you only need a stub for validate because you're already handling that in `FeatmonSpec`. I don't think it explicitly _needs_ the validate method, but might be good to provide it as a stub with a comment to clarify why you aren't validating here like others do.

lib/metric-config-parser/metric_config_parser/config.py
mikewilli, line 236: We should add another one of these and another `elif` block below for `FeatmonConfig` type.

mikewilli (dismissed):
Looks good overall, but a couple suggestions to stick to the existing patterns. And can you add tests?

One thing to consider here is whether we want to add this to the `.script/validate.py` now or later. This gets called in CI to validate the config files, and I think should be a pretty straightforward addition here, but if you want to do this in a follow-up that's OK too.

I think another follow-up that would be nice is to tie this in to the metric-hub MCP.

lib/metric-config-parser/metric_config_parser/featmon.py
```diff
@@ -56,19 +59,15 @@ def nimbus_slug(self) -> str:
 class FeatmonSpec:
     """Represents a Nimbus feature monitoring config file for a single application.
 
-    The expected use is like::
+    The expected use is via ``ConfigCollection``::
 
-        FeatmonSpec.from_file(Path("featmon/firefox_desktop.toml"))
+        collection = ConfigCollection.from_github_repo()
+        for app_name, spec in collection.featmon_configs.items():
```
scholtzan: I believe `collection.featmon_configs` is a list, so `.items()` won't work here

scholtzan (approved):
Looks good to me, one issue in a comment that may need fixing.

lib/metric-config-parser/metric_config_parser/featmon.py
yashikakhurana: nice catch, yeah before that I was working with the dict

PR #1408, opened 2026-04-23
feat(nimbus): add default support for experiment segmentation by several countries

moibra05:
Adds country-based segment definitions for 16 major markets (US, JP, UK, CA, AU, DE, FR, PL, CH, IT, NL, AT, ES, IN, ID, BR, MX) on desktop, fenix, and ios

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -819,6 +819,13 @@ submission_date_column = "submission_date"
 client_id_column = "client_id"
 deprecated = false
 
+[data_sources.clients_daily]
+from_expression = "mozdata.fenix.baseline_clients_daily"
+submission_date_column = "submission_date"
+client_id_column = "client_id"
+friendly_name = "Clients Daily"
+description = "Clients Daily"
+
 [segments]
 
 [segments.new_users]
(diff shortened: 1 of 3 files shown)


yashikakhurana:
@moibra05 have you verified the codes are correct? as per the iso standard that we use

moibra05:
i cross referenced everything with this [list](https://www.iban.com/country-codes)

definitions/fenix.toml
```diff
@@ -819,6 +819,13 @@ submission_date_column = "submission_date"
 client_id_column = "client_id"
 deprecated = false
 
+[data_sources.clients_daily]
+from_expression = "mozdata.telemetry.clients_daily"
```
scholtzan: This table only has Firefox Desktop clients. This should use `mozdata.fenix.baseline_clients_daily` instead, which also has a `country` field

definitions/firefox_ios.toml
```diff
@@ -646,3 +646,121 @@ from_expression = """(
 submission_date_column = "submission_date"
 client_id_column = "client_id"
 deprecated = false
+
+[data_sources.clients_daily]
+from_expression = "mozdata.telemetry.clients_daily"
```
scholtzan:
```suggestion
from_expression = "mozdata.firefox_ios.baseline_clients_daily"
```

definitions/fenix.toml
scholtzan:
```suggestion
from_expression = "mozdata.fenix.baseline_clients_daily"
```

definitions/fenix.toml
```diff
@@ -837,6 +844,110 @@ description = """
...
+description = "Clients in US"
+select_expression = "COALESCE(LOGICAL_OR(country = 'US'), FALSE)"
+
+[segments.jp]
+data_source = "clients_daily"
+friendly_name = "Japan"
+description = "Clients in JP"
+select_expression = "COALESCE(LOGICAL_OR(country = 'JP'), FALSE)"
+
+[segments.uk]
```
scholtzan: maybe we should be consistent with the naming here and use "GB"?

definitions/fenix.toml
```diff
@@ -837,6 +844,110 @@ description = """
...
+[segments.us]
+data_source = "clients_daily"
+friendly_name = "United States"
+description = "Clients in US"
+select_expression = "COALESCE(LOGICAL_OR(country = 'US'), FALSE)"
+
+[segments.jp]
+data_source = "clients_daily"
+friendly_name = "Japan"
+description = "Clients in JP"
```
scholtzan: My preference would be to use the full country name in the description

scholtzan (approved):

definitions/fenix.toml
```diff
@@ -819,6 +819,13 @@ submission_date_column = "submission_date"
 client_id_column = "client_id"
 deprecated = false
 
+[data_sources.clients_daily]
+from_expression = "mozdata.fenix.baseline_clients_daily"
+submission_date_column = "submission_date"
+client_id_column = "client_id"
+friendly_name = "Clients Daily"
+description = "Clients Daily"
+
```
mikewilli, line 828: Not a big deal, but we shouldn't need regular data sources for segments (since they have their own `[segments.data_sources]` section), so we might want to come back and clean this up if we're not adding it for a particular use-case, just to avoid potential for confusion. Same for the iOS one.

PR #1411, opened 2026-04-27
Add Jetstream config for optimized-set-to-default-infobar

rfambro:
Adds a custom Jetstream config for the `optimized-set-to-default-infobar` experiment.

## What this adds
- **Built-in metrics listed for weekly/overall**: `is_default_browser_glean` (Glean-native default-browser parallel to the auto-applied `is_default_browser`), `new_profile_retained_week4` (4-week new-profile retention), `repeat_first_month_user` (early retention)
- **Custom segments by profile age** (Day 0-7, Day 7-28, Day 28+) — derived from `mozdata.telemetry.clients_last_seen` via `DATE_DIFF(submission_date, first_seen_date, DAY)`

Auto-applied metrics (DAU, search, retention via `retained`, `is_default_browser`, etc.) are not re-listed and continue to run with their default statistics.

## Experiment context
- Nimbus: https://experimenter.services.mozilla.com/nimbus/optimized-set-to-default-infobar/summary/
- Brief: [Optimize set to default infobar experience](https://docs.google.com/document/d/1z3g7kSC68WNRgO1ghBtXKtF_42rxwyueSClVx7yN50c/edit)
- Eng: OMC-1559 · DO: DO-2135 · FXE: FXE-1581

### DIFF
--- a/jetstream/optimized-set-to-default-infobar.toml
+++ b/jetstream/optimized-set-to-default-infobar.toml
@@ -8,11 +8,13 @@ segments = [
 
 [metrics]
 weekly = [
+    "is_default_browser",
     "is_default_browser_glean",
     "new_profile_retained_week4",
     "repeat_first_month_user",
 ]
 overall = [
+    "is_default_browser",
     "is_default_browser_glean",
     "new_profile_retained_week4",
     "repeat_first_month_user",
--- a/jetstream/optimized-set-to-default-infobar.toml
+++ b/jetstream/optimized-set-to-default-infobar.toml
@@ -7,23 +7,13 @@ segments = [
 
 
 [metrics]
-# Auto-applied via jetstream/defaults/firefox_desktop.toml -- not re-listed:
-# active_hours, ad_clicks, days_of_use, retained, uri_count, search_count,
-# qualified_cumulative_days_of_use, is_default_browser,
-# client_level_daily_active_users_v2 (and overall also adds organic_search_count,
-# searches_with_ads, tagged_search_count, tagged_follow_on_search_count).
-
-weekly = [
-    "new_profile_retained_week4",
-    "repeat_first_month_user",
-]
-overall = [
-    "new_profile_retained_week4",
-    "repeat_first_month_user",
-]
-
-[metrics.new_profile_retained_week4.statistics.binomial]
-[metrics.repeat_first_month_user.statistics.binomial]
+# All metrics for this experiment are auto-applied via
+# jetstream/defaults/firefox_desktop.toml: active_hours, ad_clicks, days_of_use,
+# retained, uri_count, search_count, qualified_cumulative_days_of_use,
+# is_default_browser, client_level_daily_active_users_v2 (and overall also adds
+# organic_search_count, searches_with_ads, tagged_search_count,
+# tagged_follow_on_search_count). Retention is captured via the auto-applied
+# `retained` metric (per analysis window, anchored to enrollment).
 
 
 [segments]
--- a/jetstream/optimized-set-to-default-infobar.toml
+++ b/jetstream/optimized-set-to-default-infobar.toml
@@ -7,56 +7,56 @@ segments = [
 
 
 [metrics]
+# Auto-applied via jetstream/defaults/firefox_desktop.toml -- not re-listed:
+# active_hours, ad_clicks, days_of_use, retained, uri_count, search_count,
+# qualified_cumulative_days_of_use, is_default_browser,
+# client_level_daily_active_users_v2 (and overall also adds organic_search_count,
+# searches_with_ads, tagged_search_count, tagged_follow_on_search_count).
+
 weekly = [
-    "is_default_browser",
-    "is_default_browser_glean",
     "new_profile_retained_week4",
     "repeat_first_month_user",
-    "client_level_daily_active_users_v2",
-    "retained",
-    "days_of_use",
-    "uri_count",
-    "active_hours",
-    "search_count",
-    "ad_clicks",
-    "qualified_cumulative_days_of_use",
 ]
 overall = [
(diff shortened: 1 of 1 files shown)


jetstream/optimized-set-to-default-infobar.toml
```diff
@@ -0,0 +1,77 @@
...
+    "new_profile_retained_week4",
+    "repeat_first_month_user",
+    "client_level_daily_active_users_v2",
+    "retained",
+    "days_of_use",
+    "uri_count",
+    "active_hours",
+    "search_count",
+    "ad_clicks",
+    "qualified_cumulative_days_of_use",
```
mikewilli: You don't need to include these metrics since they are included as defaults.

jetstream/optimized-set-to-default-infobar.toml
```diff
@@ -0,0 +1,77 @@
...
+    "tagged_search_count",
+    "tagged_follow_on_search_count",
+]
+
+
+[segments]
+
+[segments.profile_age_0_to_7_days]
+select_expression = """COALESCE(ANY_VALUE(profile_age_bucket) = '0_to_7_days', FALSE)"""
+data_source = "clients_last_seen_with_profile_age"
```
mikewilli, line 28: Consider adding a `friendly_name` and `description` to each of the segments.

jetstream/optimized-set-to-default-infobar.toml
```diff
@@ -0,0 +1,77 @@
...
+window_start = 0
+window_end = 0
+
+
+[segments.data_sources]
+
+[segments.data_sources.clients_last_seen_with_profile_age]
+from_expression = """(
+    SELECT
+        *,
```
mikewilli: Not a big deal, but usually best practice is to select only the fields you need. In this case, you should be able to select only the `client_id, profile_group_id, submission_date` (though I haven't looked at the table schema to confirm these are the exact names).

jetstream/optimized-set-to-default-infobar.toml
```diff
@@ -0,0 +1,77 @@
...
+]
+
+
+[segments]
+
+[segments.profile_age_0_to_7_days]
+select_expression = """COALESCE(ANY_VALUE(profile_age_bucket) = '0_to_7_days', FALSE)"""
+data_source = "clients_last_seen_with_profile_age"
+window_start = 0
+window_end = 0
```
mikewilli, line 30: Why are all of these using window start/end of 0? I need to refresh my knowledge here but this doesn't seem right to me on first glance.

jetstream/optimized-set-to-default-infobar.toml
```diff
@@ -0,0 +1,77 @@
...
+    "profile_age_more_than_28_days",
+]
+
+
+[metrics]
+weekly = [
+    "is_default_browser",
+    "is_default_browser_glean",
+    "new_profile_retained_week4",
+    "repeat_first_month_user",
```
mikewilli: If you take a look at the CI error, you'll see something like `ValueError: No statistical treatment defined for metric 'is_default_browser_glean'`. Each metric to be analyzed for an experiment needs to have at least one statistic defined.

mikewilli (changes requested):

jetstream/optimized-set-to-default-infobar.toml
rfambro: Good catch — dropped `is_default_browser_glean` in the latest commit. We'll just rely on `is_default_browser` from the desktop defaults.

jetstream/optimized-set-to-default-infobar.toml
rfambro: You're right — dropped the auto-applied metrics from both `weekly` and `overall` to follow the convention from #1404. Removed `is_default_browser`, `client_level_daily_active_users_v2`, `retained`, `days_of_use`, `uri_count`, `active_hours`, `search_count`, `ad_clicks`, `qualified_cumulative_days_of_use`, `organic_search_count`, `searches_with_ads`, `tagged_search_count`, and `tagged_follow_on_search_count`. Left `new_profile_retained_week4` and `repeat_first_month_user` (not in defaults) and added explicit `binomial` stat blocks for both — they would have hit the same CI error as `is_default_browser_glean` otherwise (first uses in metric-hub, no upstream stats).

jetstream/optimized-set-to-default-infobar.toml
rfambro: Done — added `friendly_name` and `description` to each segment, matching the pattern from #1404.

jetstream/optimized-set-to-default-infobar.toml
rfambro: Good question — you caught a real issue. The intent is to bucket clients by profile age **at enrollment**, but the original `select_expression` used `submission_date`, which means the bucket would shift over the analysis window (a "0–7 days" client at enrollment would roll into "7–28 days" by week 2).

Rewrote to anchor on `e.enrollment_date` and `first_seen_date` directly, mirroring the pattern from #1404:

```toml
select_expression = """COALESCE(
    DATE_DIFF(MIN(e.enrollment_date), ANY_VALUE(first_seen_date), DAY) BETWEEN 0 AND 6,
    FALSE
)"""
```

Window stays `0/0` since `first_seen_date` is fixed per client — we only need a single-day snapshot to grab it.

jetstream/optimized-set-to-default-infobar.toml
rfambro: Tightened — now selects `client_id, profile_group_id, submission_date, first_seen_date` only. Kept `first_seen_date` since the segment bucket logic depends on it.

mikewilli (approved):

PR #1412, opened 2026-04-27
[ci rerun-skip] feat(nimbus): add 3dr metric for fenix and ios

moibra05:
Adds 3-day retention ingestion for mobile experiments

### DIFF
--- a/definitions/fenix.toml
+++ b/definitions/fenix.toml
@@ -114,6 +114,12 @@ description = """
 owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
 deprecated = false
 
+[metrics.active_in_last_3_days]
+select_expression = "COALESCE(MIN(days_since_active), 30) < 3"
+data_source = "fenix_active_users_view"
+friendly_name = "3 Days Retention"
+description = "Records whether a client submitted any pings (i.e. used Firefox) on any of the last 3 days."
+
 [metrics.user_reports_site_issue_count]
 data_source = "events"
 select_expression = """
@@ -817,6 +823,7 @@ from_expression = """(
 )"""
 submission_date_column = "submission_date"
 client_id_column = "client_id"
+experiments_column_type = "native"
 deprecated = false
 
 [data_sources.clients_daily]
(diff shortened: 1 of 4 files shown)


definitions/fenix.toml
```diff
@@ -114,6 +114,12 @@ description = """
 owner = ["bochocki@mozilla.com", "firefox-kpi@mozilla.com"]
 deprecated = false
 
+[metrics.active_in_last_3_days]
+select_expression = "COALESCE(MIN(days_since_active), 30) < 3"
+data_source = "fenix_active_users_view"
```
mikewilli, line 119: This data source doesn't specify the `experiments_column_type`, which gives us a few extra metrics and filters that are useful when applicable. In this case, this data source would be `"native"`. Can you add that to both the fenix and ios versions of the data source?

mikewilli (changes requested):
The metrics look great! One thing I'd request here, and then you can decide if you want to combine making these the default with this change, or follow up with a separate PR for that.

mikewilli:
@moibra05 Please update the PR title so it starts with `[ci skip-rerun]`. Since the PR is changing the default metrics, we don't want to trigger a rerun for all live experiments.

There's more info on the README here if you're curious. https://github.com/mozilla/metric-hub#experiment-analysis-considerations

jetstream/defaults/fenix.toml
```diff
@@ -5,6 +5,7 @@ daily = [
     "active_hours",
     "total_uri_count",
     "search_count",
+    "active_in_last_3_days"
```
mikewilli: Trailing comma, just to keep with the existing formatting:
```suggestion
    "active_in_last_3_days",
```

jetstream/defaults/firefox_ios.toml
```diff
@@ -5,6 +5,7 @@ daily = [
     "active_hours",
     "search_count",
     "total_uri_count",
+    "active_in_last_3_days"
```
mikewilli:
```suggestion
    "active_in_last_3_days",
```

mikewilli (approved):
Please just make the updates to include the trailing commas, otherwise LGTM

moibra05:
should the pr title be `[ci skip-rerun]` or `[ci rerun-skip]`? There looks to be an inconsistency between what the README says versus the gh actions bot

PR #1415, opened 2026-04-29
fix(nimbus): rename event_stream to events_stream, add SourceType enum, add ratios to FeatureSpec

yashikakhurana:
Because

- `events_stream` matches the actual BigQuery table name; `event_stream` was inconsistent
- `ratios` was referenced in the bigquery-etl query template but had no way to be defined in TOML configs
- The source type was validated against a plain frozenset with no type safety

This commit

- Renames `event_stream` → `events_stream` in `VALID_SOURCE_TYPES`, `SourceTableSpec` default, and `firefox_desktop.toml`
- Adds `SourceType` enum (`metrics`, `events_stream`, `clients_daily`) for type-safe validation
- Adds `ratios: list[list[str]]` to `FeatureSpec` so TOML configs can define metric ratios
- Updates `FeatmonSpec.from_dict` to populate `ratios` from config
- Bumps version to `2026.4.4`

Related: EXP-6732

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/featmon.py
+++ b/lib/metric-config-parser/metric_config_parser/featmon.py
@@ -5,7 +5,7 @@ checkout, with one TOML file per application (e.g. ``firefox_desktop.toml``).
 
 Each file defines:
 - ``data_sources``: BigQuery tables to query, each with a ``type`` of
-  ``"metrics"``, ``"event_stream"``, or ``"clients_daily"``
+  ``"metrics"``, ``"events_stream"``, or ``"clients_daily"``
 - ``features``: Nimbus features and the metrics to collect per source table
 
 The ``dataset`` is derived from the filename stem (e.g. ``firefox_desktop.toml``
@@ -16,15 +16,23 @@ and are accessible via ``ConfigCollection.featmon_configs``.
 """
 
 from collections.abc import Mapping
+from enum import StrEnum
 from pathlib import Path
 from typing import Any
 
 import attr
 import toml
 
-VALID_SOURCE_TYPES = frozenset({"metrics", "event_stream", "clients_daily"})
-
 FEATMON_DIR = "featmon"
+
+
+class DataSourceType(StrEnum):
+    """Valid BigQuery source table types for featmon configs."""
+
+    METRICS = "metrics"
+    EVENTS_STREAM = "events_stream"
+    DEPRECATED_EVENT_STREAM = "event_stream"  # deprecated alias, use events_stream
+    CLIENTS_DAILY = "clients_daily"
 
 
 @attr.s(auto_attribs=True)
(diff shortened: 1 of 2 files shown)


yashikakhurana:
@mikewilli The integration tests clone firefox_desktop.toml from the main branch on GitHub rather than from this PR branch, so they were loading type = "event_stream" while the updated validator now only accepts events_stream — causing the failures.

To unblock CI, I've added event_stream as a deprecated alias in SourceType. Once this PR merges, main will have type = "events_stream" in the TOML and the alias can be removed in a follow-up. Both the parser change and the TOML change are in this PR, so they land together.

lib/metric-config-parser/metric_config_parser/featmon.py
```diff
@@ -16,24 +16,32 @@
...
 
 import attr
 import toml
 
-VALID_SOURCE_TYPES = frozenset({"metrics", "event_stream", "clients_daily"})
-
 FEATMON_DIR = "featmon"
 
 
+class SourceType(StrEnum):
```
mikewilli: Maybe `DataSourceType`?

lib/metric-config-parser/metric_config_parser/featmon.py
```diff
@@ -16,24 +16,32 @@
...
+    """Valid BigQuery source table types for featmon configs."""
+
+    METRICS = "metrics"
+    EVENTS_STREAM = "events_stream"
+    EVENT_STREAM = "event_stream"  # deprecated alias, use events_stream
```
mikewilli: Perhaps we should include something more obvious than a comment, like `DEPRECATED_` prefix?
```suggestion
    DEPRECATED_EVENT_STREAM = "event_stream"  # deprecated alias, use events_stream
```

lib/metric-config-parser/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mozilla-metric-config-parser"
-version = "2026.4.3"
+version = "2026.4.4"
```
mikewilli: You might consider waiting on this. You can have an immediately follow-up to remove the deprecated event_stream option that then pushes a new version.

mikewilli (approved):
Made a couple comments, but everything here looks good. 

Did you have something changed in the `jetstream/` dir at some point? I'm wondering why the CI put in some messages about experiment analysis, that should only run when `jetstream/**` stuff is changed.

Also, the integration test issue seems like a bad thing that we should fix (separately), I'll have to think about it. If you're happy with landing this and then updating again afterwards to remove the deprecated event_stream option, then I don't think we need to block this. Thanks!

yashikakhurana:
> Made a couple comments, but everything here looks good.
> 
> Did you have something changed in the `jetstream/` dir at some point? I'm wondering why the CI put in some messages about experiment analysis, that should only run when `jetstream/**` stuff is changed.
> 
> Also, the integration test issue seems like a bad thing that we should fix (separately), I'll have to think about it. If you're happy with landing this and then updating again afterwards to remove the deprecated event_stream option, then I don't think we need to block this. Thanks!

yeah my bad did added some changes which were not required

PR #1417, opened 2026-04-29
fix(nimbus): remove deprecated event_stream alias and bump version to 2026.4.4

yashikakhurana:
Because

- `DEPRECATED_EVENT_STREAM = "event_stream"` was added as a temporary alias in #1415 to keep CI passing while `firefox_desktop.toml` still had `type = "event_stream"` on main
- Now that #1415 has merged, `firefox_desktop.toml` uses `events_stream` throughout and the alias is no longer needed

This commit

- Removes `DEPRECATED_EVENT_STREAM` from `DataSourceType`
- Bumps version to `2026.4.4`

Related: EXP-6732


mikewilli (approved):
Beautiful :shipit:

PR #1424, opened 2026-05-05
Add Jetstream config for set-to-default-prompt-style-spotlight-v2

rfambro:
Adds a custom Jetstream config for the `set-to-default-prompt-style-spotlight-v2` experiment.

## What this adds
- 3 profile-age-at-enrollment segments computed from `clients_last_seen.first_seen_date`:
  - `profile_age_0_to_7_days`
  - `profile_age_7_to_28_days`
  - `profile_age_more_than_28_days`
- No metric overrides — all success/guardrail metrics come from the auto-applied `firefox_desktop.toml` defaults.

Mirrors the segment definitions from the merged `optimized-set-to-default-infobar.toml` config so spotlight-v2 results can be cut by the same profile-age cohorts.

## Experiment context
- Nimbus: https://experimenter.services.mozilla.com/nimbus/set-to-default-prompt-style-spotlight-v2

### DIFF
--- a/jetstream/set-to-default-prompt-style-spotlight-v2.toml
+++ b/jetstream/set-to-default-prompt-style-spotlight-v2.toml
@@ -1,0 +1,65 @@
+[experiment]
+segments = [
+    "profile_age_0_to_7_days",
+    "profile_age_7_to_28_days",
+    "profile_age_more_than_28_days",
+]
+
+
+[metrics]
+# All metrics for this experiment are auto-applied via
+# jetstream/defaults/firefox_desktop.toml: active_hours, ad_clicks, days_of_use,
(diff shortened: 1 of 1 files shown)


mikewilli (approved):
This looks the same as the last one, so looks good to me.

By the way, if this is going to be a common set of segments, it would be better to add them to `definitions/firefox_desktop.toml`. That would make them available in Experimenter so you can add them before launching the experiment, and then you wouldn't need to add a config file at all.

You'll still need a config file for this and any other live experiments that don't already have these segments, but it would help you for any future experiments that have yet to launch. If you want to do that, I'd suggest creating a separate PR for that specifically.

PR #1434, opened 2026-05-08
add custom config for false end date

jaredsnyder:
I want to add a false end date to see the results with out actually ending this experiment:
https://experimenter.services.mozilla.com/nimbus/sections-in-germany/summary/

### DIFF
--- a/jetstream/sections-in-germany.toml
+++ b/jetstream/sections-in-germany.toml
@@ -1,2 +1,2 @@
 [experiment]
-end_date = "2026-05-08"
+end_date = "2026-05-07"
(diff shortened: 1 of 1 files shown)


mikewilli (approved):
Yep, that's all there is to it!

Keep in mind that jetstream runs `overall` analysis on the day after the end date (so that the end date itself is complete). If this lands as-is with today's date as the end date, it will run all the daily/weekly analysis up to today, and then do one more run tomorrow that finishes up the last day/week(maybe, if 8 May would complete a full week)/overall. This is fine and maybe intended, but dates are tricky so just wanted to clarify.

jaredsnyder:
Thanks, I did want it to run the overall for today so this is very helpful!  Adjusted the date to be yesterday

PR #1437, opened 2026-05-11
fix(ci): validate script includes enrollments for segments

mikewilli:
Because
- segments are only used for experimentation
- segments can reference fields in the enrollments with `e.` prefix

This commit
- adds a mock enrollments CTE for segment validation

I also added `venv` to the gitignore and some perf timing to the validate script.

### DIFF
--- a/.github/workflows/validate-metrics.yml
+++ b/.github/workflows/validate-metrics.yml
@@ -4,13 +4,9 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - 'definitions/**'
   push:
     branches:
       - main
-    paths:
-      - 'definitions/**'
 
 jobs:
   changed:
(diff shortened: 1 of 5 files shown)


scholtzan (approved):

.github/workflows/validate-metrics.yml
```diff
@@ -4,13 +4,9 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - 'definitions/**'
   push:
     branches:
       - main
-    paths:
-      - 'definitions/**'
```
mikewilli, line 13: @scholtzan It looks like we could remove the paths filter here and in `validate-jetstream`, and leave the `changed` job to determine whether the validation job runs or is skipped. If it's skipped with the `job.if` check, then it reports success and can be required. If it's skipped with the path filter here then it reports no status and will never pass as a required check.

scholtzan: Yeah, we could try that

mikewilli: Alright, I'll remove path filters for validate-jetstream in addition to this one and merge.

PR #1441, opened 2026-05-12
Additional config for split view feature usage metric

BryanBaird:
This adds a small change for the ["split-view-onboarding" experiment](https://experimenter.services.mozilla.com/nimbus/split-view-onboarding/results/?analysis_basis=enrollments&reference_branch=control&segment=all) analysis to add one additional metric for feature-specific usage. Since the experiment is already complete, the intent is to have one additional rerun to calculate the effect on this metric at the daily, weekly, and overall levels.

My understanding from repo conventions and the [experimenter.info online documentation](https://experimenter.info/data-analysis/jetstream/configuration) is that a basic additive change like this should not necessarily require a formal review and approval before merging to trigger a rerun. That said, I would appreciate a quick review since:

1. This is my first time submitting a config change for an experiment, and
2. I'm not sure if the use of the default `metrics` data source is unduly inefficient for this calculation.

### DIFF
--- a/jetstream/split-view-onboarding.toml
+++ b/jetstream/split-view-onboarding.toml
@@ -10,7 +10,7 @@ weekly = ["split_view_usage"]
 overall = ["split_view_usage"]
 
 [metrics.split_view_usage]
-select_expression = "mozfun.map.extract_keyed_scalar_sum(metrics.labeled_counter.splitview_uri_count) > 0"
+select_expression = "COALESCE(LOGICAL_OR(mozfun.map.extract_keyed_scalar_sum(metrics.labeled_counter.splitview_uri_count) > 0), FALSE)"
 data_source = "metrics"
 friendly_name = "Split-view feature DAU"
 description = "The number of unique daily clients loading at least one URI in a split view tab pane (either side)."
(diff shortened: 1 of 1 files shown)


jetstream/split-view-onboarding.toml
```diff
@@ -0,0 +1,18 @@
...
+[metrics]
+
+daily = ["split_view_usage"]
+
+weekly = ["split_view_usage"]
+
+overall = ["split_view_usage"]
+
+[metrics.split_view_usage]
+select_expression = "mozfun.map.extract_keyed_scalar_sum(metrics.labeled_counter.splitview_uri_count) > 0"
```
scholtzan: Validation is failing as this needs to be an aggregation. I think you'd probably just have to wrap it like:

```suggestion
select_expression = "COALESCE(LOGICAL_OR(mozfun.map.extract_keyed_scalar_sum(metrics.labeled_counter.splitview_uri_count) > 0), FALSE)"
```

(depending on how you want to handle null values).

Metric definitions describe an aggregation of activities or measurements for a specific entity, e.g. a client

BryanBaird: That makes sense! I wasn't sure if the aggregation needed to happen in the data source definition, pre-aggregate by dates, etc. But this does make sense as an implementation that would accurately describe the desired logic when aggregating:
 - by client
 - across arbitrary date windows

scholtzan (approved):

PR #1450, opened 2026-05-15
Add custom configs for Smart Shortcuts v4

mlplyler:
Adds Jetstream custom configs for smart-shortcuts-v4p1 and smart-shortcuts-v4p2.

For both we are adding urlbar-clicks.

For v4p1, I want the experimenter UI metrics plus this additional one.

For v4p2, we are adding some metrics when none existed before.

### DIFF
--- a/jetstream/smart-shortcuts-v4p1.toml
+++ b/jetstream/smart-shortcuts-v4p1.toml
@@ -1,0 +1,91 @@
+[experiment]
+reference_branch = "control"
+
+# overwrite nonessential metrics with no-op
+
+[metrics]
+
+overall = ['urlbar_topsite_clicks']
+
+[metrics.organic_pocket_clicks]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+statistics = { bootstrap_mean = {} }
+
+[metrics.newtab_searches]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+statistics = { bootstrap_mean = {} }
+
+[metrics.newtab_pocket_enabled]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+statistics = { bootstrap_mean = {} }
+
+[metrics.newtab_sponsored_pocket_stories_enabled]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+statistics = { bootstrap_mean = {} }
+
+[metrics.organic_pocket_impressions]
+select_expression = 'SUM(0)'
+data_source = "noop_source"
+statistics = { bootstrap_mean = {} }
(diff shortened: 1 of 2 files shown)


jetstream/smart-shortcuts-v4p1.toml
```diff
@@ -0,0 +1,11 @@
+[experiment]
+reference_branch = "control"
```
mikewilli: FYI this shouldn't hurt anything, but also isn't necessary.

jetstream/smart-shortcuts-v4p2.toml
```diff
@@ -0,0 +1,134 @@
+[experiment]
+reference_branch = "control"
```
mikewilli: (same here)

jetstream/smart-shortcuts-v4p2.toml
```diff
@@ -0,0 +1,134 @@
...
+
+[metrics.newtab_ad_click_rate.statistics.bootstrap_mean]
+[metrics.newtab_ad_click_rate.statistics.deciles]
+
+[metrics.newtab_homepage_enabled.statistics.binomial]
+[metrics.newtab_engagement.statistics.binomial]
+[metrics.newtab_any_searches.statistics.binomial]
+[metrics.newtab_gt4_searches.statistics.binomial]
+
+[metrics.sponsored_tile_clicks.statistics.sum]
```
mikewilli: I don't think the `sum` statistic is normally used for experiment analysis, the mean you have on all these seems sufficient and more typical. If you want this for some separate analysis you can certainly keep it, but it won't be displayed in Experimenter.

mikewilli: (same comment applies to the other instances of `sum`, I won't make a separate comment for each one)

jetstream/smart-shortcuts-v4p1.toml
```diff
@@ -0,0 +1,11 @@
...
+
+[metrics]
+overall = ['urlbar_topsite_clicks']
+
+[metrics.urlbar_topsite_clicks]
+select_expression = "SUM(IF(product_result_type = 'top_site', urlbar_clicks, 0))"
+data_source = "urlbar_events_daily_engagement_by_product_result_type_v1"
+friendly_name = "Urlbar topsite clicks"
```
mikewilli, line 7: I always suggest including a `description` for new metrics, but it's not necessary if you'd rather not. It would be displayed in Experimenter though if you include it.

mikewilli (approved):
No blockers, but I left some suggestions.

PR #1454, opened 2026-05-18
update false end date

jaredsnyder:
Update false end date for the `sections-in-germany` experiment so that it runs today


mikewilli (approved):
I don't know the dates for this experiment, but just making sure you're intending to be 6 days off from last week and not 7 (in case there's a weekly analysis boundary).

jaredsnyder:
Yep, the main thing being fixed here is excluding last Wednesday which is enough to get us another week.  Thank you for flagging though!

PR #1455, opened 2026-05-20
[ci rerun-skip] Add improved Retention metric for Desktop Defaults

BryanBaird:
This pull request is the first in a series of code and definition changes for key default metrics in Desktop Experiments related to DAU and retention. These changes are intended to better align the recommended metrics with DS team current best practices, and make it easier to interpret metrics when answering questions about the effects of treatments on overall app usage.

As a first change, this PR adds an updated version of the flagship `Retained` with one that explicitly looks for DAU-qualifying activity. The intention is to have this version run alongside the existing version (that counts *any* ping as retained), for a brief period to evaluate the effective difference between the new and old metrics. Then, a separate code change will more fully retire the old `Retained` version from the defaults, alongside changes to other DAU and activity metrics, as part of a communicated rollout.

[ci rerun-skip]

### DIFF
--- a/jetstream/defaults/firefox_desktop.toml
+++ b/jetstream/defaults/firefox_desktop.toml
@@ -2,7 +2,7 @@
 
 [metrics]
 daily = [
-  "retained",
+  "retained_dau",
   "unenroll",
   "client_level_daily_active_users_v2",
   "active_hours",
@@ -14,7 +14,7 @@ weekly = [
   "active_hours",
   "ad_clicks",
   "days_of_use",
-  "retained",
+  "retained_dau",
   "uri_count",
   "search_count",
   "qualified_cumulative_days_of_use",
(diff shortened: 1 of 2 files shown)


BryanBaird:
> ⚠️ Changing default metrics will cause all experiments that are currently live to get rerun after this change has been merged. This may come with a substantial cost. To skip reruns, add `[ci rerun-skip]` to the PR message.

I imagine this is pretty important to get right for this change, or else it might cause a lot of expensive reruns with how many experiments this would apply to.

jetstream/defaults/firefox_desktop.toml
```diff
@@ -94,6 +96,7 @@ period = "preenrollment_week"
 
 
 [metrics.retained.statistics.binomial]
+[metrics.retained_dau.statistics.binomial]
```
mikewilli, line 100: I'd suggest separating these with a blank line so it doesn't get confused for being two statistics on the same metric.

mikewilli (commented):
Seems correct, but suggested a small formatting tweak to avoid potential confusion. I'd also suggest putting `[ci rerun-skip]` at the beginning of the PR title so it's more obvious (though it should work either way).

mikewilli (approved):

PR #1460, opened 2026-05-27
Add Jetstream config for pdf-organizesplit-promo-for-default-users-5050

rfambro:
Adds a custom Jetstream config for the `pdf-organizesplit-promo-for-default-users-5050` experiment (live since 2026-05-07).

## What this adds

**9 custom binomial metrics** built on the `pdfjs.organize.action` labeled counter ([Glean dictionary](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/pdfjs_organize_action)). One combined top-line plus one per action label, so we can both ship on the headline ("did the callout get any Organize & Split adoption?") and diagnose which actions actually moved:

- `pdfjs_organize_action_any` — any of the 8 labels triggered
- Per-label binaries: `_copy`, `_cut`, `_delete`, `_move`, `_paste`, `_export_selected` (the Split action), `_save`, `_merge`

Both `weekly` and `overall`. Pattern mirrors the existing `pdf_highlight` / `pdf_engagement` definitions in `definitions/firefox_desktop.toml` (binomial, `data_source = "metrics"`, `mozfun.map.get_key` on the labeled counter).

Note: there is no separate `pdfjs.split.action` metric — `pdfjs.organize.action` covers both Organize and Split actions (`export_selected` is the Split-to-new-PDF action; `merge` is merge-with-another-PDF).

**2 profile-age-at-enrollment segments** ("new vs established" cut from the brief):

- `profile_age_0_to_28_days` — custom (mirrors the segment from `set-to-default-guidance-notification-gif.toml`)
- `profile_age_more_than_28_days` — built-in (from `definitions/firefox_desktop.toml` via #1427)

## Experiment context

- Nimbus: https://experimenter.services.mozilla.com/nimbus/pdf-organizesplit-promo-for-default-users-5050
- Brief: PDF Organize & Split Feature Callout — Default Handlers (1A of a coordinated two-experiment initiative)

### DIFF
--- a/jetstream/pdf-organizesplit-promo-for-default-users-5050.toml
+++ b/jetstream/pdf-organizesplit-promo-for-default-users-5050.toml
@@ -1,0 +1,147 @@
+[experiment]
+segments = [
+    "profile_age_0_to_28_days",
+    "profile_age_more_than_28_days",
+]
+
+
+[metrics]
+
+weekly = [
+    "pdfjs_organize_action_any",
+    "pdfjs_organize_action_copy",
+    "pdfjs_organize_action_cut",
+    "pdfjs_organize_action_delete",
+    "pdfjs_organize_action_move",
+    "pdfjs_organize_action_paste",
+    "pdfjs_organize_action_export_selected",
+    "pdfjs_organize_action_save",
+    "pdfjs_organize_action_merge",
+]
+
+overall = [
+    "pdfjs_organize_action_any",
+    "pdfjs_organize_action_copy",
+    "pdfjs_organize_action_cut",
+    "pdfjs_organize_action_delete",
+    "pdfjs_organize_action_move",
+    "pdfjs_organize_action_paste",
+    "pdfjs_organize_action_export_selected",
+    "pdfjs_organize_action_save",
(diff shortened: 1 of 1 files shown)


jetstream/pdf-organizesplit-promo-for-default-users-5050.toml
```diff
@@ -0,0 +1,147 @@
...
+
+[segments.data_sources.clients_last_seen_with_profile_age]
+from_expression = """(
+    SELECT
+        client_id,
+        profile_group_id,
+        submission_date,
+        first_seen_date
+    FROM mozdata.telemetry.clients_last_seen
+)"""
```
mikewilli, line 147: I thought we added this and other similar segments to the root definitions so that you could re-use them. Is that right? If so, I think you only need to reference it at the top `segments` list instead of recreating it here. If not, then we probably should! 

I'm guessing that we added some others, but not this one specifically.

rfambro: I believe that we added some others. But this one is worth adding to the root definitions as well.

PR #1494, opened 2026-06-11
[ci rerun-skip] Remove 28 day preenrollment for Firefox Desktop defaults

BryanBaird:
By all current indication, nothing consumes the 28 day pre-enrollment data for Desktop, making it a lot of excess overhead to run by default on all Desktop experiments. It appears these were included as part of an investigation into alternate CUPED windows for defaults, before the 7-day convention was chosen. 

The functionality is still available, if there is cause to add it for a specific experiment config or future default metric. But if there are not any _current_ consumers, I believe we should remove this from the list of defaults calculated for all Desktop experiments.

[ci rerun-skip]

### DIFF
--- a/jetstream/defaults/firefox_desktop.toml
+++ b/jetstream/defaults/firefox_desktop.toml
@@ -54,21 +54,6 @@ preenrollment_weekly = [
   "client_level_daily_active_users_v2",
 ]
 
-preenrollment_days28 = [
-  "active_hours",
-  "ad_clicks",
-  "days_of_use",
-  "organic_search_count",
-  "search_count",
-  "searches_with_ads",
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

mikewilli (approved):
This looks good, but want to go ahead and do this for the other apps while you're at it?

BryanBaird:
> This looks good, but want to go ahead and do this for the other apps while you're at it?

Cleaning up old open PRs: @mikewilli I went ahead and removed for the 4 other defaults where I saw this option in place. Good for re-approval and merge?

mikewilli (approved):
Beautiful, thanks for following up!

PR #1502, opened 2026-06-16
feat: add new optional experiment params for forcing reruns

mikewilli:
These fields will be sent by the Experimenter API and used by Jetstream to determine if a full rerun should be triggered instead of normal analysis behavior.

cc @yashikakhurana


scholtzan (approved):

yashikakhurana (approved):
Thank you @mikewilli looks good to me

PR #1515, opened 2026-06-25
Don't set `skip_default_metrics = true` for `all-users-day-5-reengage…

ncalexan:
…ment-push`.


ncalexan:
This doesn't seem to do anything locally (https://github.com/mozilla/jetstream/issues/3277) but I see errors with these default metrics in Experimenter.  Let's see if this fixes it!

scholtzan (approved):

PR #1516, opened 2026-06-26
update to linear model mean and move pocket_newtab outcomes to newtab…

ilanasegall:
…_do_no_harm


jaredsnyder (approved):
Also strips out the 10-shortcuts-per-row experiment, and leaves pocket_newtab for another time.  Other than that does what it says, lgtm

PR #1525, opened 2026-07-02
Allow Outcomes to specify metrics at other analysis window levels

BryanBaird:
Addresses an [old proposal document from Jetstream](https://github.com/mozilla/metric-hub/pull/new/outcomes-per-metric-analysis-periods).

Currently, `outcomes` are hard-coded to only calculate included metrics/statistics for the `overall` and `weekly` analysis windows. This change is intended to allow the same specification format and approach that is used in Defaults and experiment-specific custom configs, which allows for listing out which metrics to calculate at the other valid AnalysisWindow values, especially `preenrollment_weekly` (needed for CUPED/covariate adjustment) and `daily` (preferred analysis window for some metrics meant to be reused across experiments.

This is a **Claude-drafted** implementation of the change, so it will need some review from someone who has a better understanding of how the metric parsing fits together with downstream steps in Jetstream, mozanalysis, and experimenter. The included tests might also not be sufficient for meaningful unit/integration testing purposes. But for what it's worth, my surface-level review of Claude's output and reasoning does suggest that the changes really can be implemented purely from within the metric-config-parsing here, instead of changes needed in those other repos.

Note also that for backwards-compatibility, this approach will fall back to the current `overall`/`weekly` intervals for *config* that does not specify interval lists. It may be preferable to instead fall back per-metric.


mikewilli (commented):
Looks reasonable on first pass, but I'll take a closer look after the work week.

One thing that probably should be updated as well is the example config files like https://github.com/mozilla/metric-hub/blob/main/jetstream/outcomes/firefox_desktop/example_config.toml.example (not sure if people use these example configs much, but we should make sure they are up-to-date).

lib/metric-config-parser/metric_config_parser/tests/test_config.py
```diff
@@ -1015,6 +1015,36 @@ def test_entity_from_path_recognizes_featmon(self, tmp_path):
...
+            from_expression = "1"
+            """
+        )
+        (outcome_dir / "good_outcome.toml").write_text(config)
+
+        entity = entity_from_path(outcome_dir / "good_outcome.toml")
+
+        assert isinstance(entity, Outcome)
+        assert [m.name for m in entity.spec.overall] == ["rocks_mined"]
+        assert [m.name for m in entity.spec.days28] == ["rocks_mined"]
```
mikewilli, line 1046: Maybe assert that `weekly` is `[]` since the default behavior was weekly+overall?

mikewilli (approved):
Small suggestion but LGTM thanks!

mikewilli:
Oh and one other thing -- update the package version in pyproject so that a new version is published with this change.

lib/metric-config-parser/metric_config_parser/tests/test_config.py
BryanBaird: Added for weekly and the other periods as well. Though if I understand your intent correctly, the more interesting test is actually the one here, that shows how the list resolves after the config specification is merged with the default fallback behavior: https://github.com/mozilla/metric-hub/blob/cee7b671730381d4e113bc07e091125e795fe766/lib/metric-config-parser/metric_config_parser/tests/test_outcomes.py#L385-L387

BryanBaird:
@mikewilli Doing a bit of cleanup -- applied your recommended changes, including cleaning up the example TOML files to use preenrollment/CUPED adjustments as a motivating example for why other analysis windows are especially useful for Outcomes.

If this still looks good to you, we can merge and close this out.

lib/metric-config-parser/metric_config_parser/tests/test_config.py
mikewilli: Ah yea that covers it. Doesn't hurt to have it here too though since you've already made the change.

mikewilli (approved):

PR #1541, opened 2026-07-10
metric-config-parser: allow experiments_column_type "json"

danielkberry:
Adds `"json"` to metric-config-parser's `DataSource.EXPERIMENT_COLUMN_TYPES` (the config-validation allowlist), so metric-hub configs can set `experiments_column_type = "json"` for data sources whose `experiments` column is a top-level BigQuery `JSON` map (e.g. `*.events_stream` tables).

## Why
The existing `"events_stream"` type gates enrollment-day events on per-event `event_extra`, which only nimbus event rows carry — so it drops **all** enrollment-day metric events for events_stream sources (observed: control exposure read **16.2%** vs **56.4%** ground truth on a Fenix sponsored-tiles experiment). The new `"json"` type reads the client-level `experiments` map instead.

This is the **config-parser half** of the fix. The SQL-generation half is [mozilla/mozanalysis#542](https://github.com/mozilla/mozanalysis/pull/542) — metric-config-parser only validates/parses the type; mozanalysis builds the day-0 attribution guard from the client-level map.

## Changes
- add `"json"` to `EXPERIMENT_COLUMN_TYPES` + docstring
- tests: accepts `"json"`, rejects unknown types
- bump version `2026.6.1` → `2026.7.0` to publish the parser to PyPI on merge

## Sequencing
Prerequisite for the metric-hub data-source switch ([#1540](https://github.com/mozilla/metric-hub/pull/1540)) and for the `jetstream` image (which needs both this parser release and the mozanalysis release) to accept/generate the `"json"` type.

### DIFF
--- a/lib/metric-config-parser/metric_config_parser/data_source.py
+++ b/lib/metric-config-parser/metric_config_parser/data_source.py
@@ -72,6 +72,12 @@ class DataSource:
               string.
             * 'events_stream': There is an ``experiment`` within a JSON
               column ``event_extra``. ``branch`` is in the same column.
+            * 'json': There is a top-level ``experiments`` column of
+              BigQuery ``JSON`` type, which is an (experiment_slug:str ->
+              struct) map with a ``branch`` field. This is the client-level
+              experiments map carried on every row of ``*.events_stream``
+              tables; unlike 'events_stream' it attributes metric events on
+              the enrollment day.
             * None: There is no ``experiments`` column, so skip the
               sanity checks that rely on it. We'll also be unable to
               filter out pre-enrollment data from day 0 in the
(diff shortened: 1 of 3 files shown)


mikewilli:
I think this is covered by the `"events_stream"` option for `experiments_column_type`. If something's missing from that option, I'd suggest we update that instead of adding a new one. Or if this option was not discoverable, then maybe it's just missing some docs.

(edit to add: sorry I didn't mean to close the PR with my comment, that was a misclick. Leaving this open to discussion still.)

mikewilli:
Ok I looked a little more closely and see that you have some concerns with the existing `events_stream` type, let me give this a proper review after the work week.

PR #1632, opened 2026-08-19
[ci rerun-skip] feat: add new outcomes to win10 experiments

eddchen-mozilla:


mikewilli (approved):
Changes look good and the relevant [Jetstream changes](https://github.com/mozilla/jetstream/pull/3302) have landed as well, which is what allows for `outcomes` to be specified in this way.

PR #1616, opened 2026-08-18
Add and use Fenix notification definitions

ncalexan:

PR #1634, opened 2026-08-20
Add Fenix notification outcome.

ncalexan:
This is #1616 plus the outcome.


mikewilli (approved):

mikewilli:
Oops I meant to leave a comment -- I only looked at the outcome since you said the other parts are the same as #1616, but looks good.

PR #1646, opened 2026-09-01
[ci rerun-skip] feat(config): add mean_lm no-clip default as additional stat to default metrics

mikewilli:
- adds an additional `LinearModelMeanNoClip` statistic to the metrics that currently have `mean_lm` configured
- my IDE sorted all the daily/weekly/etc lists when I saved and I decided that wasn't a bad idea

~Blocked by https://github.com/mozilla/jetstream/pull/3310 because we don't want to land this until the results can be differentiated in the output.~ this is landed now

### DIFF
--- a/jetstream/defaults/fenix.toml
+++ b/jetstream/defaults/fenix.toml
@@ -1,59 +1,59 @@
 [metrics]
 daily = [
-    "retained",
-    "client_level_daily_active_users_v2",
     "active_hours",
-    "total_uri_count",
-    "search_count",
     "active_in_last_3_days",
+    "client_level_daily_active_users_v2",
+    "retained",
+    "search_count",
+    "total_uri_count",
 ]
 weekly = [
-    "retained",
     "active_hours",
+    "client_level_daily_active_users_v2",
+    "days_of_use",
+    "retained",
     "search_count",
     "serp_ad_clicks",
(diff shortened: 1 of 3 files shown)


scholtzan (approved):

doomsuckle:
Some context here for @danielkberry, we're putting in a version where there is no clipping by default in code and then we're going to put in the configs for where it's appropriate to keep clipping now where it's expected for topline metrics. Then we'll have an alt version we can migrate to without any visible change on the front.

doomsuckle (approved):

mikewilli:
A couple additional notes:
- I will hold off on merging this until at least next week, pending resolution of the BigQuery issues we're currently experiencing. While this only affects the stats and therefore the cluster side of execution, I don't want to add additional strain to the system right now.
- Experimenter [only ingests](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/jetstream/models.py#L39-L51) the `mean_lm` statistic, so this shouldn't affect anything on the UI. We can change the code at that link if we want, but I think the better option is to use this to evaluate behind the scenes (i.e., by querying the results tables directly), and then when we're satisfied that the default should be no clipping, change the default for the `mean_lm` stat and delete `mean_lm_noclip` altogether.

PR #1663, opened 2026-09-09
[ci rerun-skip] Move 3-day retention out of Desktop default and into Onboarding outcome

BryanBaird:
Follow-up from https://github.com/mozilla/metric-hub/pull/1525. 

Now that the option is available to calculate daily metrics in Outcomes, this continues the previous implementation plan to move "3 Day Retention" out of the list of defaults for _all_ experiments, since it was designed to be used for new profiles specifically.

This also includes some metadata updates to the metric friendly UI name and description to explain this nuance.

**Not** in scope of this change:
1. Any corresponding changes in mobile -- this is Desktop only.
2. Any Experimenter UI changes that might be required to account for 3 Day Retention no longer being a KPI on every experiment.

[ci rerun-skip]

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -1580,14 +1580,25 @@ description = """
 [metrics.active_in_last_3_days]
 select_expression = "COALESCE(MIN(days_since_desktop_active), 30) < 3"
 data_source = "firefox_desktop_baseline_active_users_view"
-friendly_name = "3 Days Retention"
-description = "Records whether a client submitted any pings (i.e. used Firefox) on any of the last 3 days."
+friendly_name = "First 3 Days Retention (Glean)"
+description = """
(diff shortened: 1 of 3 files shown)


BryanBaird:
Note that I think this will fail until Jetstream picks up the package version bump from PR #1525.

mikewilli:
>Note that I think this will fail until Jetstream picks up the package version bump from PR https://github.com/mozilla/metric-hub/pull/1525.

Yea that seems right to me. I'll go ahead and do that and rerun CI here when I'm done.

mikewilli (approved):
As you said, we'll have to wait for the checks to pass once I finish updating jetstream, but this change makes sense to me. There should probably be some communication coordinated with this change so that teams running onboarding/new-user experiments know to add this outcome to get this metric.

PR #1666, opened 2026-09-11
[ci rerun-skip] Retire older DAU-related metrics from Desktop defaults

BryanBaird:
Following up from the changes started in https://github.com/mozilla/metric-hub/pull/1501 and https://github.com/mozilla/metric-hub/pull/1455, this PR retires the other previously-used Desktop defaults that measure versions of DAU and Retention that are now superseded by `DAU per 1,000 clients` and the improved `Retained (DAU)`.

Specifically removed are:

- `client_level_daily_active_users_v2`
- `days_of_use`
- `qualified_cumulative_days_of_use`
- `retained`

"Retirement" here means that these other metrics are removed from the list of Desktop defaults. Their metadata descriptions are also updated with an explanatory note on their changes and rationales, and the `deprecated` value is set to `True`. Many of these metrics have both Legacy and Glean versions defined, if not used in the current set of defaults -- these explanatory notes point to the corresponding Legacy/Glean versions of the new metrics.

Definitions remain available for experiments and outcomes that reference them explicitly, and this change is not meant to invalidate any results where these are already calculated.

### Note for reviewers on non-experiment use cases
I do not fully understand non-experiment implications for changing metrics like this, that might be referenced in other OpMon or Looker contexts. My understanding is that changing the metadata will not break any of these downstream applications, and at most will cause a `deprecated` banner to appear in some limited cases.

### Note on sequencing
The Experimenter UI has some hard-coded references to these values still in place, in terms of how they are presented in the Results UI. Those display references will need to be updated to avoid displaying extra errors as experiment results start coming in without values for these metrics.

More details are ticketed in Jira here: https://mozilla-hub.atlassian.net/browse/EXP-7516 

### Note on communications
Because this change will affect the default values and interpretation of metrics that are highly visible, we will want to message widely to the affected audiences. These new metadata descriptions are meant to be a bit more complete in order to explain some of the context directly in-place, but more proactive outreach is also warranted.

### Scope limit
This change is **Desktop-only.** As with other recent changes, a separate mobile equivalent will need to be applied separately.

[ci rerun-skip]

### DIFF
--- a/definitions/firefox_desktop.toml
+++ b/definitions/firefox_desktop.toml
@@ -362,13 +362,25 @@ deprecated = true
 data_source = "clients_daily"
 select_expression = "COUNT(submission_date)"
 friendly_name = "Days of use"
-description = "The number of days in the interval that each client sent a main ping."
+description = """
+    The number of days in the interval that each client sent a main ping.
+
+    Note: As of September 2026, this metric is being deprecated in favor of "DAU per 1,000 clients", the streamlined
+    DAU metric now recommended by Data Science for measuring DAU effects in experiments.
+"""
+deprecated = true
 
 [metrics.days_of_use_glean]
 data_source = "baseline_clients_daily"
 select_expression = "COUNT(submission_date)"
 friendly_name = "Days of use in Glean"
-description = "The number of days in the interval that each client sent a baseline ping."
+description = """
+    The number of days in the interval that each client sent a baseline ping.
+
+    Note: As of September 2026, this metric is being deprecated in favor of "DAU per 1,000 clients (Glean)", the streamlined
+    DAU metric now recommended by Data Science for measuring DAU effects in experiments.
+"""
+deprecated = true
 
 [metrics.qualified_cumulative_days_of_use]
 data_source = "clients_daily"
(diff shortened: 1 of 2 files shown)


mikewilli (commented):
Here are my thoughts on deprecating all these metrics: removing the metrics from the default list will ensure they don't get computed for experiments unless someone explicitly adds them in a custom config (which seems very unlikely). If they are no longer useful for any purpose, then deprecating them is the right move. If they could be useful outside of experimentation, then I'd omit the deprecated flag, and adjust the comment to note that they are no longer recommended by DS for experiment analysis as of Sept 2026, but may be useful otherwise (for [x,y,z], if you have examples of where to use them).

And just so you know, I don't think the deprecated flag is anything more than a recommendation, but that's ok, the flag + description change should get the point across, if deprecation is what we want.

mikewilli:
>The Experimenter UI has some hard-coded references to these values still in place, in terms of how they are presented in the Results UI. Those display references will need to be updated to avoid displaying extra errors as experiment results start coming in without values for these metrics.

Is this true? I know there are hard-coded references, but I think the UI will just happily ignore them if they are not present. I think we can remove [this code](https://github.com/mozilla/experimenter/blob/efe6bad3f889a7834f47f286b9895d6c5bbc0889/experimenter/experimenter/jetstream/results_manager.py#L342-L360), but it shouldn't cause any problems if we don't.

/cc @yashikakhurana Do you know of any places where the UI might error if an expected KPI metric is not there?

mikewilli:
I've noticed that certain KPI metrics get treated as "not yet available" in the UI, i.e., they appear in the UI with a message even if they don't appear in the results. Screenshot below:
image

I'm guessing this means we'll just need to adjust a list somewhere in Experimenter of which metrics are expected for KPIs. If a particular metric is not going to be in the results, then Experimenter shouldn't show a message that implies it should/will be there.

PR #1668, opened 2026-09-11
FormAutofill related metrics

rolf-moz:
Implemented as part of https://mozilla-hub.atlassian.net/browse/FXCM-2589

Adds per-client form autofill metrics and attaches them to two desktop
field-detection experiments.

https://experimenter.services.mozilla.com/nimbus/form-autofill-ml/summary/ just completed on Nightly and a new version just launched https://experimenter.services.mozilla.com/nimbus/form-auto-fill-field-detection-improvements-v2/summary/

These experiments will hopefully soon graduate to Beta and Release

The goal is to track field fills and modifications from Form Autofill. Overall form submissions are included also as a guard metric to check for unexpected changes.

### Metrics

All 8 are defined in `definitions/firefox_desktop.toml` (category `autofill`) so
both experiments — and any future autofill work — share one implementation:

| Metric | Meaning |
|---|---|
| `form_autofill_address_fields_filled` | gross auto-filled address fields per client |
| `form_autofill_address_fields_modified` | of those, how many the user then edited |
| `form_autofill_address_fields_net_kept` | auto-filled and left untouched |
| `form_autofill_credit_card_fields_*` | same as above, but for credit card forms |
| `form_autofill_address_form_submissions` | distinct address forms submitted |
| `form_autofill_credit_card_form_submissions` | distinct credit card forms submitted |

`filled` = `modified` + `net_kept` by construction.

### Data source

One new data source, `form_autofill_forms`: one row per client / day / form
interaction (the shared `flow` uuid), covering address and credit card forms,
carrying per-field detected / auto-filled / modified / net-kept counts plus a
submitted flag.

All 8 metrics share it deliberately. `event_extra` is ~104 GB/day on desktop
`events_stream` and jetstream scans once per data source, so splitting field
metrics from submission metrics would have tripled the scan for identical
results.

### Definition decisions worth reviewing

**1. "Detected" includes Fathom, not just regex/ML.** A `detected_*` event
reports per field how the type was established: `false` = not on the form,
`true` = explicit `autocomplete` attribute, `0` = regex heuristic, a number =
Fathom confidence, `ml` = ML classifier. A field counts when Firefox *inferred*
the type — regex, Fathom or ML — excluding `autocomplete`-attribute fields,
whose type needs no inference and so is unaffected by a classifier change.

This matters for credit cards. Measured on 2026-09-09, gross fields per client:

| | regex+ML only | incl. Fathom |
|---|---|---|
| address | 0.2281 | 0.2289 (+0.35%) |
| credit card | 0.0133 | **0.0257 (+93%)** |

Fathom is roughly half of all auto-filled CC fields, so excluding it would
halve the CC metrics. Address is unaffected either way.

**2. Edited sub-fields are folded onto the reported field.** `filled_modified_*`
events use a finer vocabulary than the detected/filled events. `tel-national`
etc. roll up to `tel`, `cc-given-name`/`cc-family-name`/`cc-additional-name` to
`cc-name`, and `country-name` to `country`.

`address-housenumber` (~2% of address modifications) needed a judgment call: it
is a sub-input of a split street address, and in those flows autofill had
filled `address-line1` 56% of the time and `street-address` 20%, but both only
1%. So the edit is credited to both — naming a field that was not auto-filled
is a no-op, since `modified` is only read for fields that were detected *and*
filled. Effect: address modified per client 0.0231 → 0.0234, net kept 0.2119 →
0.2116, gross unchanged. `address-level3` (3 flows/day) has no counterpart key
and is left unmapped.

### Experiments

| slug | status | enrollment | randomization |
|---|---|---|---|
| `form-autofill-ml` | Complete (ended 2026-09-11) | 2026-08-27 → 2026-09-03 | group_id |
| `form-auto-fill-field-detection-improvements-v2` | Live (2026-09-11) | — | group_id |

Both get all 8 metrics on `weekly`, `overall` and `preenrollment_weekly`, with
`linear_model_mean`, `linear_model_mean_no_clip` and `deciles`. These are
zero-inflated long-tailed counts, so the clipped and unclipped means plus the
full distribution let a volume cap be chosen downstream rather than baked in.

Both randomize on `group_id`, so analysis resolves to `profile_group_id`
automatically and no `analysis_unit` override is needed; `form_autofill_forms`
exposes `profile_group_id` alongside `client_id` and glean/legacy ids.
`profile_group_id` is non-null on 99.93% of autofill events.

### Rerun and cost notes

- **Deliberately not using `[ci rerun-skip]`.** `form-autofill-ml` ended
  2026-09-11 and has no results for these metrics, so the rerun that a new
  custom config triggers is the point. It ended within 3 days, so Experimenter
  should auto-refresh.
- The jetstream data-processing estimate may trip the high-threshold warning,
  given `event_extra` volume across three windows. If it needs trimming, the
  cheapest lever is dropping `preenrollment_weekly` and the two
  `covariate_adjustment` blocks that depend on it.

### Validation

- `entity.validate()` passes for the definitions file and both jetstream
  configs; all 8 metrics resolve to `form_autofill_forms` with the expected
  statistics in all three windows.
- The CI-rendered `validation_query.sql` for all 8 metrics dry-runs clean.
- `filled = modified + net_kept` holds exactly — 0 violations across 17.1M
  clients, both form types.
- 0 duplicate rows on the data source's declared grain, so the `SUM` metrics
  cannot double count.

Note for reviewers: an outer join at the top of the `from_expression` breaks
BigQuery's pushdown of the caller's `submission_date` filter to the
`events_stream` partition column, which makes every metric query fail the
mandatory-partition-filter check. The submitted-forms side is therefore unioned
in rather than joined — there's a comment in the SQL to that effect.

### Caveats

- The data source is pre-aggregated, so `experiments_column_type = "none"`:
  day-0 pre-enrollment events can't be filtered out.
- A form interaction spanning midnight is split across two daily rows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jetstream/form-auto-fill-field-detection-improvements-v2.toml
+++ b/jetstream/form-auto-fill-field-detection-improvements-v2.toml
@@ -1,0 +1,117 @@
+# form-auto-fill-field-detection-improvements-v2
+#
+# Live field-detection experiment (control | locall-ml | local-ml-two-head),
+# launched 2026-09-11. Follow-up to form-autofill-ml.
+#
+# The eight form autofill metrics and the `form_autofill_forms` data source they
+# read from are defined in definitions/firefox_desktop.toml, so this experiment
+# and form-autofill-ml share one implementation.
+#
+# Nimbus randomizes this experiment on group_id, so analysis runs on
+# profile_group_id; no analysis_unit override is needed because
+# `form_autofill_forms` exposes profile_group_id alongside client_id.
+#
+# enrollment_period and end_date are deliberately omitted so the Nimbus values
+# apply (per metric-hub review feedback on PR #1401).
+
+[metrics]
+weekly = [
+  "form_autofill_address_fields_filled",
+  "form_autofill_address_fields_modified",
+  "form_autofill_address_fields_net_kept",
+  "form_autofill_credit_card_fields_filled",
+  "form_autofill_credit_card_fields_modified",
+  "form_autofill_credit_card_fields_net_kept",
+  "form_autofill_address_form_submissions",
+  "form_autofill_credit_card_form_submissions",
+]
+overall = [
+  "form_autofill_address_fields_filled",
+  "form_autofill_address_fields_modified",
+  "form_autofill_address_fields_net_kept",
+  "form_autofill_credit_card_fields_filled",
+  "form_autofill_credit_card_fields_modified",
+  "form_autofill_credit_card_fields_net_kept",
+  "form_autofill_address_form_submissions",
+  "form_autofill_credit_card_form_submissions",
+]
+# Requested for the pre-enrollment week as well, so the statistics below can
+# covariate-adjust on each group's own prior autofill behaviour.
+preenrollment_weekly = [
+  "form_autofill_address_fields_filled",
+  "form_autofill_address_fields_modified",
+  "form_autofill_address_fields_net_kept",
+  "form_autofill_credit_card_fields_filled",
+  "form_autofill_credit_card_fields_modified",
+  "form_autofill_credit_card_fields_net_kept",
+  "form_autofill_address_form_submissions",
+  "form_autofill_credit_card_form_submissions",
+]
+
+# Statistics: these are zero-inflated, long-tailed per-client counts, so report
+# the mean both with the default 0.5% top clip and unclipped, plus the full
+# distribution so a volume cap can be chosen downstream rather than baked in.
+
+[metrics.form_autofill_address_fields_filled.statistics.linear_model_mean]
+[metrics.form_autofill_address_fields_filled.statistics.linear_model_mean.covariate_adjustment]
+period = "preenrollment_week"
+[metrics.form_autofill_address_fields_filled.statistics.linear_model_mean_no_clip]
+[metrics.form_autofill_address_fields_filled.statistics.linear_model_mean_no_clip.covariate_adjustment]
+period = "preenrollment_week"
+[metrics.form_autofill_address_fields_filled.statistics.deciles]
+
+[metrics.form_autofill_address_fields_modified.statistics.linear_model_mean]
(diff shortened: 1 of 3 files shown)


jetstream/form-auto-fill-field-detection-improvements-v2.toml
```diff
@@ -0,0 +1,117 @@
...
+  "form_autofill_address_fields_net_kept",
+  "form_autofill_credit_card_fields_filled",
+  "form_autofill_credit_card_fields_modified",
+  "form_autofill_credit_card_fields_net_kept",
+  "form_autofill_address_form_submissions",
+  "form_autofill_credit_card_form_submissions",
+]
+# Requested for the pre-enrollment week as well, so the statistics below can
+# covariate-adjust on each group's own prior autofill behaviour.
+preenrollment_weekly = [
```
rolf-moz, line 40: @irrationalagent is this pre-enrollment stuff needed?  I don't know what' standard

hamzahamidi (approved):
Reviewed the aggregation and experiment definitions. The per-form aggregation and submit-only union avoid double counting, and the validation checks are green.

hamzahamidi (commented):
No actionable code issue found in the aggregation or experiment definitions. Validation checks are green.

DimiDL (approved):

jetstream/form-auto-fill-field-detection-improvements-v2.toml
irrationalagent: its good to have because it can reduce the noise in the results (basically allows jetstream to baseline the results on any random preexisting differences in the metrics that might exist between the branches).

irrationalagent (approved):
looks good. lets see how the dry run on the datasource goes, sometimes these queries can be quite heavy. but the team ha improved the query orchestration somewhat recently so im optimistic this will work

